PR #40, opened 2015-01-21
39 re add webrtc ready callback

johache:
I added ```AdapterJS.WebRTCReadyCallback``` which will be called whenever the WebRTC API is ready (on any browser, not only for the plugin).
See comments in the code for more details.

The blocking loop waiting for the plugin to be ready is still present in case people don't want to use this callback.

I added a link to the existing plugin documentation where this will be documented (today hopefully)

This needs to be tested against SkylinkJS before merging


johache:
wanted to merge into dev, not master

PR #41, opened 2015-01-21
39 re add webrtc ready callback

johache:
I added AdapterJS.WebRTCReadyCallback which will be called whenever the WebRTC API is ready (on any browser, not only for the plugin).
See comments in the code for more details.

The blocking loop waiting for the plugin to be ready is still present in case people don't want to use this callback.

I added a link to the existing plugin documentation where this will be documented (today hopefully)

This needs to be tested against SkylinkJS before merging.


serrynaimo:
@johache I think it makes sense to make the callback name AdapterJS.onwebrtcready - That's usually how these things look in JS. Please insert a snippet into the README.md that explains that this is where people should get their code going.

johache:
@serrynaimo review again:
Changed callback name to onwebrtcready, plus style refactoring.

PR #44, opened 2015-02-02
Fix AdapterJS.WebRTCPlugin.pluginNeededButNotInstalledCb not being overridable

jamix2:
As per the comments in `adapter.js`, `AdapterJS.WebRTCPlugin.pluginNeededButNotInstalledCb` should be overridable. However, it wasn't because `AdapterJS.WebRTCPlugin.pluginNeededButNotInstalledCb` was being assigned unconditionally further down the code. This pull request fixes this.

PR #45, opened 2015-02-06
Fix AdapterJS.WebRTCPlugin.pluginNeededButNotInstalledCb not being overridable

jamix2:
As per the comments in `adapter.js`, `AdapterJS.WebRTCPlugin.pluginNeededButNotInstalledCb` should be overridable. However, it wasn't because `AdapterJS.WebRTCPlugin.pluginNeededButNotInstalledCb` was being assigned unconditionally further down the code.

This is an updated version of #44 that uses `development` as the target branch.


johache:
Validated and auto-merged

PR #76, opened 2015-05-12
AJS-189: changing integration process of AJS: call immediatly if webr…

johache:
IMPORTANT: @serrynaimo , @letchoo , what do you think

I realized that if AJS was imported before the onwebrtcready callback was defined, Chrome and firefox would not work. The callback is always done before being overridden by the app.

I suggest we display AdapterJS.webrtcready publicly, and change the integration process from

```
// You should define this callback BEFORE you import AdapterJS, otherwise it might be called before your callback is ready.
 
var AdapterJS = AdapterJS || {}; // this is to make sure AdapterJS is defined before setting the callback
AdapterJS.onwebrtcready = function(isUsingPlugin) {
    // The WebRTC API is ready.
    //isUsingPlugin: true is the WebRTC plugin is being used, false otherwise
    getUserMedia(constraints, successCb, failCb);
};
```

To

```
var AdapterJS = AdapterJS || {};
if (AdapterJS.webrtcready) {
    getUserMedia(constraints, successCb, failCb);
} else {
  AdapterJS.onwebrtcready = function(isUsingPlugin) {
    // The WebRTC API is ready.
    //isUsingPlugin: true is the WebRTC plugin is being used, false otherwise
    getUserMedia(constraints, successCb, failCb);
  }
}
```

### DIFF
--- a/source/adapter.js
+++ b/source/adapter.js
@@ -90,7 +90,7 @@ AdapterJS.WebRTCPlugin.pluginState = AdapterJS.WebRTCPlugin.PLUGIN_STATES.NONE;
 
 // True is AdapterJS.onwebrtcready was already called, false otherwise
 // Used to make sure AdapterJS.onwebrtcready is only called once
-AdapterJS.onwebrtcreadyDone = false;
+AdapterJS.webrtcready = false;
 
 // Log levels for the plugin. 
 // To be set by calling AdapterJS.WebRTCPlugin.setLogLevel
@@ -150,8 +150,8 @@ __TemWebRTCReady0 = function () {
 };
 
 AdapterJS.maybeThroughWebRTCReady = function() {
-  if (!AdapterJS.onwebrtcreadyDone) {
-    AdapterJS.onwebrtcreadyDone = true;
+  if (!AdapterJS.webrtcready) {
+    AdapterJS.webrtcready = true;
 
     if (typeof(AdapterJS.onwebrtcready) === 'function') {
       AdapterJS.onwebrtcready(AdapterJS.WebRTCPlugin.plugin !== null);


serrynaimo:
```
AdapterJS.webRTCReady( function(isUsingPlugin) {
    // The WebRTC API is ready.
    //isUsingPlugin: true is the WebRTC plugin is being used, false otherwise
    getUserMedia(constraints, successCb, failCb);
  });
```
and doing what you outlined above in that function?

soulhunter1987:
I sugest that you use js promises here. https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Promise Promises are supported natively in Safari >= 7.1, so you would have to implement your own fallback, but it is rather simple to do. You would anyway need it because current W3C WebRTC spec is fully based on promises and Firefox is already widely implementing it. So your init code woult look like this:
```javascript
AdapterJS.webRTCReady().then(function(isUsingPlugin) {
    // do some code when native WebRTC or plugin functionality is available
}, function() {
    // do something when no native support and plugin is not installed
});

So promise will be resolved (first callback) when pugin is installed or native support is available, and rejected (second callback) otherwise.
You can implement your own promise interface or use some existing library (which is similar to native W3R promises) as a dependency.
```

johache:
I think that's a good idea. Shouldn't need to much work either.
@serrynaimo this is more or less what you suggested, right?

serrynaimo:
Conceptually similar. If we do ship a promise lib with AdapterJS to polyfill in old browsers we should refactor a few other code snippets as well.
Otherwise my suggestion will work for now.

soulhunter1987:
A good choice might be this one: https://github.com/taylorhakes/promise-polyfill It may be included as a bower/npm dependency.

serrynaimo:
We want the developer experience for less skilled devs to be simple and not bloat AdapterJS unnecessarily so it remains fast over time. Therefor I don't like to introduce dependencies to AdapterJS and also don't need the full promise polyfill just to make this nicer code in this one occasion. Still favoring 
```
AdapterJS.webRTCReady( function(isUsingPlugin) {
    // The WebRTC API is ready.
    //isUsingPlugin: true is the WebRTC plugin is being used, false otherwise
    getUserMedia(constraints, successCb, failCb);
  });
```

johache:
agreed

PR #84, opened 2015-05-29
AdapterJS.webRTCReady event handler

letchoo:
Please check @johache

PR #85, opened 2015-06-02
0.11.0 changes

letchoo:
- Added new function AdapterJS.webRTCReady function #84 #76 
- Added screensharing functionality in getUserMedia. Invoke { video: { mediaSource: 'window' } }.

NOTE: Please do not merge this yet.


letchoo:
Ok. I've updated the readme for the getUserMedia settings for screensharing invoke. @serrynaimo please review, merge and release

johache:
@letchoo @serrynaimo 
Some of those commits should never have landed on dev. (ex: https://github.com/Temasys/AdapterJS/commit/2698802a5d419dd6b5cbb6a43e0d78660c0af47f)
I don't think AdapterJS.webRTCReady is implemented correctly in the latest version either.

I will start running the plugin tests immediately and probably push some corrections on dev.

In general, please make PR from your branch to dev.
The changes made by the plugin team need to be reviewed by the SkylinkJS team, and vice versa.

johache:
@letchoo @serrynaimo 
Is this change necessary? 
-var AdapterJS = AdapterJS || {};
+window.AdapterJS = window.AdapterJS || {};

If we change the namespacing of AJS I think we need to move to 0.11.0. anyway

source/adapter.js
```diff
@@ -1,7 +1,12 @@
 // Adapter's interface.
-var AdapterJS = AdapterJS || {};
+window.AdapterJS = window.AdapterJS || {};
```
johache: @letchoo @serrynaimo 
Is this change necessary? 
If we change the namespacing of AJS I think we need to move to 0.11.0. anyway

publish/adapter.debug.js
```diff
@@ -118,11 +123,6 @@ AdapterJS.WebRTCPlugin.WaitForPluginReady = null;
 // This methid will use an interval to wait for the plugin to be ready.
 AdapterJS.WebRTCPlugin.callWhenPluginReady = null;
 
-// This function will be called if the plugin is needed (browser different
-// from Chrome or Firefox), but the plugin is not installed.
-// Override it according to your application logic.
-AdapterJS.WebRTCPlugin.pluginNeededButNotInstalledCb = null;
```
johache: Can we do 
AdapterJS.WebRTCPlugin.pluginNeededButNotInstalledCb = AdapterJS.WebRTCPlugin.pluginNeededButNotInstalledCb || null;
? 
I like the idea of keeping the interfaces defined at the top.

source/adapter.js
```diff
@@ -254,13 +254,71 @@ AdapterJS.maybeFixConfiguration = function (pcConfig) {
...
     elem.addEventListener(evnt, func, false);
-  } else if (elem.attachEvent) {// OLD IE DOM 
+  } else if (elem.attachEvent) {// OLD IE DOM
     elem.attachEvent('on'+evnt, func);
   } else { // No much to do
     elem[evnt] = func;
   }
 };
 
+AdapterJS.renderNotificationBar = function (text, buttonText, buttonLink) {
```
johache: Any reason why this was moved here?
This function only makes sense when you're in a non-native-webrtc browser.

johache: It is also removed from its previous namespace.
That makes sense, it can be multi-purposed, but then this is another reason to move to 0.11.0

letchoo: @johache I wanted to re-use to prompt users to install the chrome web extension so I moved it there. But in the end we redirected users to the web store instead. If you prefer it in the plugin scope I can move it.

source/adapter.screenshare.js
```diff
@@ -0,0 +1,161 @@
+(function () {
+
+  'use strict';
+
+  var tempGetUserMedia = null;
+
+  if (window.navigator.mozGetUserMedia) {
```
johache: I'm uncomfortable with the fact that we check 
window.navigator.mozGetUserMedia here
and navigator.mozGetUserMedia in the basic adapter (I know it's the same)

I feel like we should either factorise the code of store the first check somewhere (which we do actually, in webrtcDetectedBrowser)

@serrynaimo Question: do we want adapter.screenshare.js to be a standalone or does it need the adapter.js base code?

source/adapter.screenshare.js
```diff
@@ -0,0 +1,161 @@
...
+    tempGetUserMedia = window.navigator.getUserMedia;
+
+    window.navigator.getUserMedia = function (constraints, successCb, failureCb) {
+      constraints = constraints || {};
+
+      if (constraints.video ? !!constraints.video.mediaSource : false) {
```
johache: Cleaner to say if (constraints.video && constraints.video.mediaSource) ?
(same comment for the identical conditions bellow)

source/adapter.screenshare.js
serrynaimo: should be a standalone for now. eventually we might merge it.

source/adapter.screenshare.js
johache: Also what is constraints.video.mediaSource == 'camera' ...
I feel like we should define what we expect from the user when they want to get a screensharing stream.
constraints.video.mediaSource == 'screen'

source/adapter.js
serrynaimo: not necessary for me... @letchoo ?

source/adapter.screenshare.js
serrynaimo: we orientate ourselves along firefox syntax where you do mediaSource = 'window'. that's why
(See http://hancke.name/webrtc/screenshare/#/28)

johache: Following my previous comment (can't remember where), I believe FF and Chrome both accepts 'screen'.
See https://www.webrtc-experiment.com/screen-sharing/
Never tested myself on FF...

I makes more sense to me to use the 'screen' keyword semantically and because it is shared by both browsers.

In any case, It does not change that of the user call mediaSource == 'camera', you should not start the screensharing.

source/adapter.js
johache: If you don't use it I think we should move it back for now.
Putting a function in a public namespace means that we have to maintain it in the future.

publish/adapter.screenshare.js
```diff
@@ -0,0 +1,1246 @@
...
+
+  } else {
+    tempGetUserMedia = window.navigator.getUserMedia;
+
+    window.navigator.getUserMedia = function (constraints, successCb, failureCb) {
+      constraints = constraints || {};
+
+      if (constraints.video ? !!constraints.video.mediaSource : false) {
+        // check if plugin is ready
+        if(AdapterJS.WebRTCPlugin.pluginState === 4) {
```
johache: Magic number !

source/adapter.screenshare.js
```diff
@@ -0,0 +1,161 @@
...
+
+  'use strict';
+
+  var tempGetUserMedia = null;
+
+  if (window.navigator.mozGetUserMedia) {
+    tempGetUserMedia = window.navigator.getUserMedia;
+
+    window.navigator.getUserMedia = function (constraints, successCb, failureCb) {
+      constraints = constraints || {};
```
johache: The spec says that constraints is not optional and not nullable.
We should not have to default constraints this way.
https://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediadevices-getusermedia

source/adapter.screenshare.js
```diff
@@ -0,0 +1,161 @@
...
+
+  } else {
+    tempGetUserMedia = window.navigator.getUserMedia;
+
+    window.navigator.getUserMedia = function (constraints, successCb, failureCb) {
+      constraints = constraints || {};
+
+      if (constraints.video ? !!constraints.video.mediaSource : false) {
+        // check if plugin is ready
+        if(AdapterJS.WebRTCPlugin.pluginState === 4) {
```
johache: Magic number !

Also, should use the functions AdapterJS.WebRTCPlugin.WaitForPluginReady and AdapterJS.WebRTCPlugin.callWhenPluginReady

source/adapter.screenshare.js
```diff
@@ -0,0 +1,161 @@
...
+            constraints.video.optional.push({
+              sourceId: AdapterJS.WebRTCPlugin.plugin.screensharingKey || 'Screensharing'
+            });
+
+            delete constraints.video.mediaSource;
+          } else {
+            throw new Error('The plugin installed does not support screensharing');
+          }
+        } else {
+          throw new Error('The plugin is currently not yet available');
```
johache: English

source/adapter.screenshare.js
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  if (window.webrtcDetectedBrowser === 'chrome') {
+    var iframe = document.createElement('iframe');
+
+    iframe.onload = function() {
+      iframe.isLoaded = true;
+    };
+
+    iframe.src = 'https://cdn.temasys.com.sg/skylink/extensions/detection-script/detectRTC.html';
```
johache: Can we move this link into a config file?

source/adapter.screenshare.js
```diff
@@ -0,0 +1,161 @@
...
+      constraints = constraints || {};
+
+      if (constraints.video ? !!constraints.video.mediaSource : false) {
+        constraints.video.mediaSource = 'window';
+        constraints.video.mozMediaSource = 'window';
+
+        tempGetUserMedia(constraints, successCb, function (error) {
+          if (error.name === 'PermissionDeniedError' && window.parent.location.protocol === 'https:') {
+            console.error(error);
+            window.location.href = 'https://cdn.temasys.com.sg/skylink/extensions/skylink-webrtc-tools.xpi';
```
johache: Can we move this link into a config file?

source/adapter.screenshare.js
```diff
@@ -0,0 +1,165 @@
...
+
+  } else {
+    baseGetUserMedia = window.navigator.getUserMedia;
+
+    window.navigator.getUserMedia = function (constraints, successCb, failureCb) {
+      constraints = constraints || {};
+
+      if (constraints.video && !!constraints.video.mediaSource) {
+        // check if plugin is ready
+        if(AdapterJS.WebRTCPlugin.pluginState === AdapterJS.WebRTCPlugin.PLUGIN_STATES.READY) {
```
johache: Should use the functions AdapterJS.WebRTCPlugin.WaitForPluginReady and AdapterJS.WebRTCPlugin.callWhenPluginReady

johache: @serrynaimo These functions are in AJS.
It means that we can either decide that this file is NOT a standalone and needs to be compile with AJS, or create a third file that contains the shared functions.
Given that we already use other function (getUserMedia defined in AJS), I will assume that this file is NOT a standalone

source/adapter.screenshare.js
```diff
@@ -0,0 +1,165 @@
...
+            constraints.video.mandatory = constraints.video.mandatory || {};
+            constraints.video.mandatory.chromeMediaSource = 'desktop';
+            constraints.video.mandatory.maxWidth = window.screen.width > 1920 ? window.screen.width : 1920;
+            constraints.video.mandatory.maxHeight = window.screen.height > 1080 ? window.screen.height : 1080;
+
+            if (sourceId) {
+              constraints.video.mandatory.chromeMediaSourceId = sourceId;
+            }
+
+            delete constraints.video.mediaSource;
```
johache: Why?
Same for lower occurrences

source/adapter.screenshare.js
```diff
@@ -0,0 +1,165 @@
...
+
+      } else {
+        baseGetUserMedia(constraints, successCb, failureCb);
+      }
+    };
+
+    window.getUserMedia = window.navigator.getUserMedia;
+
+  } else {
+    baseGetUserMedia = window.navigator.getUserMedia;
```
johache: In the basic AJS we define
navigator.getUserMedia = getUserMedia;
Therefore I don't think we should refer to window.navigator.getUserMedia here. (Even if it's technically the same)

source/adapter.screenshare.js
```diff
@@ -0,0 +1,165 @@
...
+      if (constraints.video && !!constraints.video.mediaSource) {
+        // intercepting screensharing requests
+
+        constraints.video.mediaSource = 'window';
+        constraints.video.mozMediaSource = 'window';
+
+        baseGetUserMedia(constraints, successCb, function (error) {
+          if (error.name === 'PermissionDeniedError' && window.parent.location.protocol === 'https:') {
+            console.error(error);
+            window.location.href = 'https://cdn.temasys.com.sg/skylink/extensions/skylink-webrtc-tools.xpi';
```
johache: Can we move this link into a config file?

source/adapter.screenshare.js
```diff
@@ -0,0 +1,165 @@
...
+
+  'use strict';
+
+  var baseGetUserMedia = null;
+
+  if (window.navigator.mozGetUserMedia) {
+    baseGetUserMedia = window.navigator.getUserMedia;
+
+    window.navigator.getUserMedia = function (constraints, successCb, failureCb) {
+      constraints = constraints || {};
```
johache: The spec says that constraints is not optional and not nullable.
We should not have to default constraints this way.
https://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediadevices-getusermedia

(same of lower occurrences)

source/adapter.js
serrynaimo: Let's remove the window. reference.

source/adapter.js
serrynaimo: This release will be 0.11.0 due to this interface change and the added screensharing functionality

source/adapter.screenshare.js
serrynaimo: jiraified

source/adapter.screenshare.js
letchoo: In the meeting, we decided to put the getUserMedia in the window scope.

johache:
LGTM

johache:
Passed manual and auto-test for the plugin.
The only thing that bothers me a bit is that you can give a constraints mediaSource: 'none', you will still end up with a screensharing stream

Maybe we want to be more strict about the content of mediaSource ?

letchoo:
I am fine with it. @serrynaimo what are your thoughts

serrynaimo:
Next version. Don't wanna retest again now.

PR #86, opened 2015-06-03
More updates and changes as requested

letchoo:
- Moved text into AdapterJS.Text. Did not move the one that stamps the company name.
- "window" and "screen" should be an available option in firefox.
- Throw an error if both "window" and "screen" is not selected in firefox.
- Render notification bar if extension is not yet installed.
- Clone constraints object instead of overwriting stuff.
- Stop using plugin state directly and use the AdapterJS.WaitForPluginReady


source/adapter.js
```diff
@@ -278,7 +288,7 @@ AdapterJS.addEvent = function(elem, evnt, func) {
   }
 };
 
-AdapterJS.renderNotificationBar = function (text, buttonText, buttonLink) {
+AdapterJS.renderNotificationBar = function (text, buttonText, buttonLink, openNewTab, showRefreshButton) {
```
serrynaimo: Could've used "javascript:location.reload();" as buttonLink text. Than you don't need a separate "showRefreshButton" ;)

source/adapter.screenshare.js
```diff
@@ -12,16 +27,32 @@
...
+
+        // so generally, it requires for document.readyState to be completed before the getUserMedia could be invoked.
+        // strange but this works anyway
+        var checkIfReady = setInterval(function () {
+          if (document.readyState === 'complete') {
+            clearInterval(checkIfReady);
+
+            baseGetUserMedia(updatedConstraints, successCb, function (error) {
+              if (error.name === 'PermissionDeniedError' && window.parent.location.protocol === 'https:') {
+                AdapterJS.renderNotificationBar(AdapterJS.Text.Extension.requireInstallationFF, 'Install Now',
```
serrynaimo: 'Install now' is an English string, too

PR #91, opened 2015-07-08
Add the refresh after install functionality to the plugin

sebbeaume:
The functiononality is in AdapterJS.renderNotificationBar, from line 352.


johache:
1 - Please replace tabs with double spaces 
2 - the pull request should go from your branch to development, not master

publish/adapter.debug.js
```diff
@@ -93,7 +93,7 @@ AdapterJS.WebRTCPlugin.defineWebRTCInterface = null;
 AdapterJS.WebRTCPlugin.isPluginInstalled = null;
 
  // Lets adapter.js wait until the the document is ready before injecting the plugin
-AdapterJS.WebRTCPlugin.pluginInjectionInterval = null;
+ AdapterJS.WebRTCPlugin.pluginInjectionInterval = null;
```
johache: Removed space, avoid meaningless changes.

publish/adapter.debug.js
```diff
@@ -123,7 +123,7 @@ Log outputs are prefixed in some cases.
   INFO: Information reported by the plugin.
   ERROR: Errors originating from within the plugin.
   WEBRTC: Error originating from within the libWebRTC library
-*/
+  */
```
johache: Removed space, avoid meaningless changes.

publish/adapter.debug.js
```diff
@@ -294,7 +294,14 @@ AdapterJS.addEvent = function(elem, evnt, func) {
...
 
-AdapterJS.renderNotificationBar = function (text, buttonText, buttonLink, openNewTab, displayRefreshBar) {
+
+AdapterJS.reloadPage = function()
+{
+  alert("Plug-in installed.");
+  location.reload();
+}
+
+AdapterJS.renderNotificationBar = function (text, buttonText, buttonLink, openNewTab, displayRefreshBar, func) {
```
johache: func is not clear enough

johache:
@sebbeaume for reference, have a look at the 2 commits I just added.
Question: given that we reload the plugin list, could we just do the plugin injection from onplugininstalled ? Would the plugin work in the case? That would save us a page reload.
Also please change this PR to development and not master.

sebbeaume:
@johache  I tried to use the injectPlugin function. It gives an error "getUserMedia unknown". That doesn't happen if page is reloaded.
Changing branches is impossible once the PR has been made. I have to close the request and to make a new one with the correct target.

PR #92, opened 2015-07-09
AJS-213 Add refreshing of plugin list function after plugin install

sebbeaume:


johache:
Just tested on IE:
* navigator.plugins.refresh(false) does not seem to work on my IE 11 Win 7
* I read this function is actually not supported on IE
* Once the plugin is installed, I can actually inject it and start using it however. Not page refresh required
* There might still be in solution in using what is described here:
https://msdn.microsoft.com/en-us/library/dn423948(v=vs.85).aspx
Check also AdapterJS.WebRTCPlugin.isPluginInstalled, it already does a similar job.

johache:
@serrynaimo @letchoo :
I just tested with the plugin test suite, it works fine for me.
Can you test against SkylinkJS before we merge into dev ?
The idea:
* start a session without the plugin,
* access your test page
* click on install in the top banner
* once the plugin is installed, the plugin should be automatically injected and the AdapterJS.webRTCReady and onwebrtcready should be called

johache:
@letchoo @serrynaimo Can I merge ?

PR #98, opened 2015-07-30
AJS 61 automate tests

johache:
Automated tests with Grunt, Karma, Mocha and Chai.
Not all the tests pass, but it is expected. The browsers don't fully support the spec.
The tests are not exhaustive either. They should be completed over time.

Run the tests using `grunt karma` or `grunt test` (publishes and runs karma)
The ReadMe is up to date with this.

I also git ignored the doc folder since it's auto-generated.


serrynaimo:
That's pretty badass. Cool stuff. I'd suggest we avoid temasys in filenames even in the source folder.

johache:
@serrynaimo I agree.
I think we can keep Temasys for the email addresses and to name the "Temasys WebRTC plugin", but everything else should be neutral.
I will try to rename the file today if I have the time.

johache:
@serrynaimo Renaming done.
Please note that he files now called adapter.*.js are currently not used.
They are chuncks of the "real" AJS (see @letchoo )
I think it's a really good idea if we go all the way and start generating AJS from these files.
From there, users could decide to build for only certain browsers, or even swap any given browser's adapting with their own.

PR #100, opened 2015-08-03
TWP-236 Creation of VideoElement.prop and VideoElement.behaviour. Add…

sebbeaume:
Test for muted property.

- Test that muted is boolean, default value at false.
- Test that muting video disable sounds.
- Test that a stream rendered in more than one video is muted only if all videos are muted.
- Test that a muted stream is unmuted if attached to a new video.

### DIFF
--- a/tests/unit/VideoElement.behaviour.spec.js
+++ b/tests/unit/VideoElement.behaviour.spec.js
@@ -13,135 +13,136 @@ var testItemTimeout = 4000;
 
 
 describe('VideoElement| Behaviour', function() {
-	this.timeout(testTimeout);
+  this.timeout(testTimeout);
 
-	/* Attributes */
-	var video = null;
-	var stream = null;
-	var audioTrack = null;
-	var videoTrack = null;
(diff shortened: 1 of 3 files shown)


tests/unit/VideoElement.behaviour.spec.js
```diff
@@ -0,0 +1,147 @@
...
+	before(function (done) {
+		this.timeout(testItemTimeout);
+
+		AdapterJS.webRTCReady(function() {
+			done();
+		});
+	});
+
+	/* Get User Media */
+	before(function (done) {
```
johache: Merge the two before into one

johache:
* Change the tab for double spaces ?
* remove the console.logs
* have only one before per file (each file has 2)
* you have an error in afterEach on Chrome and FF. An afterEach should NEVER fail. If it fails, it's a test
In this specific case, maybe don't use attachMediaStream with a null stream (or write a deattach function)
* two tests fail on all browsers, is this expected ?

PR #101, opened 2015-08-06
TWP-234_5 Add VideoElement.event ; Test for on play and onplaying event

sebbeaume:


tests/unit/VideoElement.event.spec.js
```diff
@@ -0,0 +1,101 @@
...
+  before(function (done) {
+    this.timeout(testItemTimeout);
+
+    AdapterJS.webRTCReady(function() {
+      done();
+    });
+  });
+
+  /* Get User Media */
+  beforeEach(function (done) {
```
johache: Magic numbers in this function

johache:
You can test more than that:
* You can test that the timestamp is a positive integer. You could even check that is makes sense with the timstamp you can get from js (with error margin, no magic number)
* You can test that the target is the video element
* You can also check currentTarget and the srcElement that should also be the video element

The test fails on my Safari, but I assume this is because I have not updated to your branch that supports the implementation of onplay.

johache:
@sebbeaume I just deleted AJS-61. You will have to create a new PR to dev... srry about that.

PR #102, opened 2015-08-11
Twp 234 5 automated test onplay onplaying

sebbeaume:


johache:
- Please remove the magic numbers,
- What does this.slow do ?
- if you de-init video in afterEach, how about setting stream to null as well?
- you do a bunch of check on onplaying. You should do it was well on onplay. 
screen shot 2015-08-12 at 10 40 33 am
- style: the opening brackets should be on the same line as the if statement
- looks like you attach the onplay event twice in your test. Note the you will likely have to link the onplay event to the plugin element in attachMediaStream

johache:
Note to self: Reading the tests looks ok, need to test when the onplay code is running properly.

johache:
My previous comment was retarded, the test should be merge whether or not it works with the plugin.
I fixed the minor issue remaining that made it fail on chrome (race issue) and merge the PR.

PR #103, opened 2015-08-13
Version 0.11.2

johache:
@letchoo @serrynaimo 
Please review for public release 0.11.2.
Most of the changes are for the new test system (you already saw when merging them on dev).
The changes added to the Adapter are:
- [bugfix] Clearing the interval that check if the plugin is installed. The interval used to trigger install callback every 1/2 sec
- Added an escape to defineWebRTCInterface so that it cannot be called more than once. Prints an error in the JS console when called more than once.


letchoo:
Looks good to me

johache:
Bumping to 0.12.0

PR #106, opened 2015-08-27
Added CDash reporter to the karma

johache:
Added cdash reporter to be able to use the tests from the plugin and publish onto CDash


letchoo:
Looks good.

PR #107, opened 2015-09-01
AJS-52 Karma test running on IE + plugin

johache:
* moved to karma-ie-launcher 0.2.0
* enabled the IE launcher on karma
* printJSON not available on IE. Replaced with JSON.stringify
* autoplay not available on IE without update. Bypass it on IE (just in case + not useful)


johache:
Re-enabled the function type check using a variable to reflect the typeof object.function
It equals 'function' everywhere except on IE + plugin where it is 'object'

johache:
ping @letchoo

johache:
There really isn't that many changes, I merged on my own ;)

PR #108, opened 2015-09-01
window.getUserMedia IE Edge fixes

letchoo:
Reference ticket AJS-228 just incase (includes information how it was fixed).

**What was fixed:**
- window.getUserMedia working in IE Edge
- window.getUserMedia should be working and not over-written in adapter.screenshare version.
- Added warnings for adapter.screenshare for browsers IE Edge and Opera that screensharing is not supported.

**Tested in**
IE Edge, IE 10, IE 11, IE 9, Safari 8.0.6, Chrome 44, Opera 31, Opera Beta 32, Firefox 41.0b5 (beta), Firefox Nightly 43.0a1

Recommended to test again :)


source/adapter.js
```diff
@@ -741,7 +741,9 @@ if (navigator.mozGetUserMedia) {
   // the minimum version still supported by adapter.
   webrtcMinimumVersion = 12;
 
-  getUserMedia = navigator.getUserMedia;
```
johache: Why not just use 
window.getUserMedia = navigator.mozGetUserMedia.bind(navigator);
navigator.getUserMedia = window.getUserMedia;
like everywhere else ???

source/adapter.screenshare.js
```diff
@@ -141,6 +141,10 @@
 
     getUserMedia = navigator.getUserMedia;
 
+  } else if (window.webrtcDetectedBrowser === 'edge') {
```
johache: I think you should use
else if (navigator.mediaDevices && navigator.userAgent.match(
    /Edge\/(\d+).(\d+)$/)) {
to keep the same system as the main adapter and not assume this variable was set.

source/adapter.js
letchoo: Refer to the JIRA ticket. It seems like Edge doesn't trigger the referenced object

```
function y () { console.log("do something"); }
var x = y;
x(); // opps error
```

source/adapter.screenshare.js
letchoo: Ok. I'm not particular how the edge should be detected as long as it works in most environments :)

letchoo:
Updated with proposed fixes. Please re-review

source/adapter.js
johache: Note: does work with binding as per your new changes

PR #114, opened 2015-09-09
Twp 274 Automated test for screensaver

sebbeaume:


johache:
@sebbeaume 
You big if seems to be an acceptable solution for plugin-only tests (http://jira.temasys.com.sg/browse/AJS-229). That's great.

I made some changes myself to save you some time.
What I did: 
- use the AJS variable ```webrtcDetectedBrowser``` instead of checking manually
- enlarging the ```if``` block to include the test describe block, 
- moved ```afterEach``` at the top, 
- added minor comments.

What's left for you:
- I want it to be easier to identify the plugin-only tests. 
Either put it in a folder or name them plugin.mytest.......spec.js
- big lol on this: 
```var startSleepTime = new Date().getTime();
      while(new Date().getTime() < startSleepTime + sleepTime){ /* do nothing */ } ```
JS is defined a single threaded language - technically -. This could possibly block you whole tab for a long time.
I'll let you check online for a better solution.
Hint: some other tests already contain the solution you are looking for.
- without testing I can tell you that ```video2.autoplay = 'autoplay';``` will fail on IE. Anyways, the plugin autoplays by default.
- you have a series of ```video.pause()``` missing a semicolon. I don't even know by what black magic the tests still pass

johache:
- the autoplay issue is still here
- check setInterval instead of setTimeout

PR #115, opened 2015-09-09
changes for phid.io

bart-jansen:


bart-jansen:
oops.. needed to go the other way

PR #117, opened 2015-09-15
AJS-231: changed <video> size copy strategy to support hidden elements

johache:
@letchoo 
We had feedbacks informing us that the <video> sizes weren't copied properly when the element was hidden of in a hidden div:

Scenario:
- have a <video> element in a hidden <div>
- set the <video> element's size to 800x600px
- attachMediaStream on it
- unhide the div

Expected:
- the plugin <object> is 800x600px

Observed:
- the plugin <object> is 0x0px

I think this should fix the issue.
clientWith and clientHeight are computed values, so when the element is hidden, they have a value of 0.

### DIFF
--- a/source/adapter.js
+++ b/source/adapter.js
@@ -1044,16 +1044,13 @@ if (navigator.mozGetUserMedia) {
 
         var height = '';
         var width = '';
-        if (element.getBoundingClientRect) {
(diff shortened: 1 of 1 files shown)


letchoo:
Looks ok to me

PR #119, opened 2015-09-16
TWP-274 test prevent screensaver

sebbeaume:


johache:
That's better. 
When using intervals you need to clean them. 
See : http://www.w3schools.com/jsref/met_win_clearinterval.asp

PR #120, opened 2015-09-17
Twp 290 onloadedmetadata automated test

sebbeaume:


tests/unit/VideoElement.event.spec.js
```diff
@@ -140,6 +140,43 @@ describe('VideoElement | EventHandler', function() {
 
   });
 
+it('VideoElement.onloadedmetadata :: emit', function (done) {
+    this.timeout(testItemTimeout);
+    video.id = 'video';
```
johache: Why setting an id here ?

tests/unit/VideoElement.event.spec.js
```diff
@@ -140,6 +140,43 @@ describe('VideoElement | EventHandler', function() {
...
+    video.onloadedmetadata = function(event) {
+      done();
+    };
+
+    video = attachMediaStream(video, stream);
+  });
+
+  it('VideoElement.onloadedmetadata :: attributes', function (done) {
+    this.timeout(testItemTimeout);
+    video.id = 'video';
```
johache: Can you make the id a random string ?
I'm afraid some day we might end up with false positive in the future.

Math.random().toString(36).slice(2);

johache:
The indentation in the adapter is a bit weird, but ok, good enough.

PR #122, opened 2015-09-21
TWP-274 Add test file

sebbeaume:


tests/unit/plugin.ScreenSaver.behaviour.spec.js
```diff
@@ -0,0 +1,187 @@
...
+    beforeEach(function (done) {
+      this.timeout(gUMTimeout);
+
+      window.navigator.getUserMedia({
+        audio: true,
+        video: true
+
+      }, function (data) {
+        stream = data;
+        videoTrack = stream.getVideoTracks()[0];
```
johache: Why get a reference to the video and audio track if you never use them ?

tests/unit/plugin.ScreenSaver.behaviour.spec.js
```diff
@@ -0,0 +1,185 @@
...
+// !!! THIS TEST ONLY APPLIES FOR PLUGIN-BASED BROWSERS !!!
+if(webrtcDetectedBrowser === 'safari' || webrtcDetectedBrowser === 'IE') {
+
+  describe('ScreenSaver | Behaviour', function() {
+    this.timeout(testTimeout);
+
+    /* Attributes */
+    var video = null;
+    var stream = null;
+    var audioTrack = null;
```
johache: Now why define never used variables ? :p

PR #124, opened 2015-09-21
Ajs 218 forward event listeners

sebbeaume:
Forwarding of the event handlers to the plugin object when it is injected in the web page.

### DIFF
--- a/source/adapter.js
+++ b/source/adapter.js
@@ -1070,17 +1070,23 @@ if (navigator.mozGetUserMedia) {
       }
       var newElement = document.getElementById(elementId);
 
+      //Event forwarding
       for(var property in element) {
         if(property.slice(0,2) == "on" && element[property] != null) {
-          if (newElement.attachEvent) {
-            //isIE
-            newElement.attachEvent(property, element[property]);
-          } else {
-            //other web browsers
-            newElement.addEventListener(property.slice(2), element[property], false); 
+          AdapterJS.addEvent(newElement, property.slice(2), element[property]);
+        }
+      }
+      if(!isIE) {
+        var nonEnumerablesToForward = ['onclick'];
+        
+        for(var event in nonEnumerablesToForward) {
+          var eventName = nonEnumerablesToForward[event]
+          if(element[eventName] != null) {
+            newElement.addEventListener(eventName.slice(2), element[eventName]);
           }
         }
       }
+
       return newElement;
     };
 
--- a/source/adapter.js
+++ b/source/adapter.js
@@ -1069,14 +1069,20 @@ if (navigator.mozGetUserMedia) {
         element.setStreamId(streamId);
       }
       var newElement = document.getElementById(elementId);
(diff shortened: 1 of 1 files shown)


source/adapter.js
```diff
@@ -1069,14 +1069,21 @@ if (navigator.mozGetUserMedia) {
...
+          if (newElement.attachEvent) {
+            //isIE
+            newElement.attachEvent(property, function(event){console.log(event)});
+          } else {
+            //other web browsers
+            newElement.addEventListener(property.slice(2), element[property], false); 
+          }
+        }
+      }
+      newElement.onclick = (element.onclick) ? element.onclick : function (arg) {};
```
johache: Why does onclick keep it own line ?

johache:
1 - PR descriptions are not optionals :p
2 - onplay is fowarded, but onclick seems not to be. Checking the console, video properties appear in 2 different colours. That might be a lead. Maybe there is a difference between inherited properties and object specific properties ?
screen shot 2015-09-22 at 2 23 54 pm 1

source/adapter.js
```diff
@@ -1069,22 +1069,24 @@ if (navigator.mozGetUserMedia) {
...
+      //Event forwarding
+      for(var property in element) {
+        if(property.slice(0,2) == "on" && element[property] != null) {
+          AdapterJS.addEvent(newElement, property.slice(2), element[property]);
+        }
       }
-      
+      if(!isIE) {
+        var nonEnumerablesToForward = ['onclick'];
+
```
sebbeaume: I couldn't find a way to iterate over the non-enumerable properties. I'm using an array containing all the properties one could be interested in forwarding.

johache:
Did you check this 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyNames
http://stackoverflow.com/questions/8024149/is-it-possible-to-get-the-non-enumerable-inherited-property-names-of-an-object

source/adapter.js
```diff
@@ -1069,22 +1069,18 @@ if (navigator.mozGetUserMedia) {
...
-          var arg = {
-            srcElement : document.getElementById(id)
-          };
-          newElement.onclick(arg);
-        };
-      }
-      
+      var elemBuff = element;
+
+      do {
```
johache: Forget about "do while" loops.
does while(var elemBuff = Object.getPrototypeOf( elemBuff )) work?
otherwise I would still prefer a for loop

johache:
Tested on IE ?

source/adapter.js
sebbeaume: while(var elemBuff = Object.getPrototypeOf( elemBuff ))  works.
I can't find a way to use for loop : we need to access the prototype of the prototype of the prototype... of the object. I don't think it's doable with a for loop in a clean way.

johache: Then go for the while loop

johache:
Good job.

PR #129, opened 2015-10-14
AJS-230 added test to pop plugin <object> in and out of the DOM

johache:
This test pops plugin <object> in and out of the DOM.
This is to check for crashes when objects are added and remove to often.


sebbeaume:
Approved

PR #132, opened 2015-10-21
Ajs 227 AdapterJS generation

johache:
This is so that the plugin config can be changed very easily (see JIRA)

* Removed the pluginInfo from AJS
* Moved it to a separate file
* compiling both on grunt publish

!!! This means that one cannot use source/adapter.js directly but has to use the compiled version !!!


letchoo:
Looks good

PR #133, opened 2015-11-03
TWP-66 Test if slice is defined. If not, use other way to forward events

sebbeaume:


johache:
Are you sure of this ?
Did you run the code ? Did you check the hints I gave you ?

sebbeaume:
Yes i ran the code. I checked the hints : i think testing both typeof === string and slice != undefined is not useful. Only testing that slice is not undefined is enough IMO.

johache:
Just checked out the code, are you REALLY sure it works ?
Did you run any test ?

johache:
LGTM, good job.

PR #134, opened 2015-11-04
AJS-237 call enableSoundTracks only when defined.

johache:
stream::enableSoundTrack is deprecated starting on plugin verion 0.8.862.
Call it only when it is defined in prevision of the function being removed.

### DIFF
--- a/source/adapter.js
+++ b/source/adapter.js
@@ -986,7 +986,9 @@ if (navigator.mozGetUserMedia) {
       if (stream === null) {
         streamId = '';
       } else {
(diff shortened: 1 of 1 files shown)


letchoo:
Looks okay to me.

johache:
If looks ok and tested, then you may merge :)

letchoo:
Tested manually with Chrome 46 <-> Safari 8.0.6 (using plugin ver 0.8.861)
Able to hear sound and video from the other party.
Test case is pass. Code passes as well.

PR #135, opened 2015-11-09
AJS-238 - Return error in failureCb instead of throwing it

letchoo:
Note that this requires to be tested manually first.
Test scripts is not working well since extensions could not be loaded in karma.
Will be skipping those tests made.

------

Related ticket: AJS-238

Please do not merge for now. To be tested first.


johache:
Please don't commit the publish folder on this kind of PR. 
It's irrelevant changes that are likely to generate merge conflicts.

Also, I think screensharing tests should be in their own file.

letchoo:
Ok noted. I'll separate the test file into `getUserMedia.screenshare.spec.js`.
Will take note and not commit the publish in the future.

letchoo:
Tested with IE 9 / 11. Tested with Safari 8.0.6. Tested with Chrome 46 and Firefox 43 and Opera 33.

Works as expected.

johache:
Did minor modifs. 
LHTM now, merging and closing.

PR #138, opened 2015-12-04
Allow multiple callbacks with webRTCReady function

rusekr:
Hello!

This is for several libraries and clients to use own webRTCReady callbacks without rewriting one.

Compatible with deprecated onwebrtcready method.

Faced today it in app where library need to set one callback with webRTCReady and app to set another. With old interface (onwebrtcready) I may create wrapper but with upgraded new no wrappers needed (:


johache:
Hi,
First of all, thanks for the PR, it's always appreciated.

I had a quick look, it looks good; 
I'm out of office until next week and won't be able to test properly until I'm back.

In the mean time, would you mind making your pull request on Temasys:development. That shouldn't create any conflict.
As much as possible, we avoid merging anything directly on master.

Thanks again, 
J-O

rusekr:
Ok! Done this in 139 (:

PR #139, opened 2015-12-05
Allow multiple callbacks with webRTCReady function

rusekr:
Hello!
As suggested in https://github.com/Temasys/AdapterJS/pull/138 done on development branch.

And without built files in publish too.


johache:
@ncurrier @serrynaimo 
Seems to work fine to me. Can I have your feedback regarding the design (by JS standard).

PR #140, opened 2015-12-17
Hack for plugin 0.8.862 version do not try to update itself

rusekr:
Hello! Made for some users stuck at this non-autoupgradable version. Tested in IE 10 Win 7 x64. Worked for me.

### DIFF
--- a/source/adapter.js
+++ b/source/adapter.js
@@ -881,6 +881,10 @@ if (navigator.mozGetUserMedia) {
     } else {
       try {
         var axo = new ActiveXObject(comName + '.' + plugName);
+        if (axo.version === "0.8.862" ) { // NOTICE: Bad version override
+          notInstalledCb();
+          return;
+        }
       } catch (e) {
         notInstalledCb();
         return;


johache:
Hi, I've just tested against IE 11, and it doesn't work for me.

It only works if the plugin actually works (in which case the auto-update works as well).
If you take a first install of 862 (which fails to init), then the call to ```new ActiveXObject(comName + '.' + plugName);``` fails and the code exits the file.
That means the code you added won't run.

Can you try it on your side and confirm my tests ?

rusekr:
I'll test in my IE11 today. In IE10 activex of plugin 0.8.862 don`t throw error , axo is object and it has version parameter.

johache:
Is this while the plugin still fails to initialise ?

rusekr:
Tested in IE10 with cleared AppData - not working. But my suggestion based on fact that plugin founds but not suggests to upgrade.

I've tested in IE10 console:

var axo = new ActiveXObject('Tem.TemWebRTCPlugin');

and axo returned without exception but == undefined - that why adapterjs not suggests to install plugin at all.

After that ive tried to create incorrect object with:

var axo = new ActiveXObject('Tema.TemWebRTCPlugin');

and it throw error. 

Thanks to your pointed to clear appdata i'll correct the path to check typeof axo and if it not object or not have version property - return notInstalledCb because wrong existed plugin === no plugin for me(: 
However it can be different callback or parametrized notInstalledCb that suggest to REinstall plugin with same link and different text but is it nessessarily?

johache:
ok, I'll let you update your patch.
I don't think it's necessary to suggest re-install, especially for a hack-fix. 
It might confuse the user more than help them.
I think using the regular notInstalledCb will be fine.

rusekr:
Catched polterGates(B): 
Even this code stucks after new ActiveXObject... event with eval it can't throw anyhing in ie10, it can`t "console.log" anything after. Some IE bug I think.. IE just ignores all after including installedCb() at the end of function..

var axo;
try {
        console.log('azaza try eval');
        axo = eval("new ActiveXObject(comName + '.' + plugName);");
        console.log('azaza', axo, typeof axo);
      } catch (e) {
        console.log('azaza catched', e);
        notInstalledCb();
        return;
      } finally {
        console.log('azaza try catch finished');
        if (typeof axo !== 'object' || !axo.version) {
          notInstalledCb();
          return;
        }
      }
installedCb()

rusekr:
IE 11 ignores code after "new ActiveXObject(comName + '.' + plugName);" too. I think of cancel trying (:

rusekr:
Closing for now because of said above

PR #153, opened 2016-02-08
added name='adapterjs-alert' to the iframe

simonelelli:
I added name='adapterjs-alert' to the iframe (Notification bar) to easily identify this element for any need as CSS manipulations.


johache:
Hi Simon,

I'm all for this modification, I think it's a pretty good idea. 
Could you however modify your PR so that it only reflects this change ? That is, without all the tab/spaces modif it currently contains ?

Thanks,
J-O

simonelelli:
Hi J-O,

thanks for your attention. I fixed the tab/spaces differences.

johache:
looks good to me, merged, thanks :)

PR #157, opened 2016-02-11
AJS-242 createiceserver throws warnings on ff43

johache:
FF43+ prints warnings when the ice candidates url in in a 'url' element.
Use 'urls' instead.
Also, printing a warning when createIceServer(s) is call, notifying the user that the functions are set to be deprecated.

### DIFF
--- a/source/adapter.js
+++ b/source/adapter.js
@@ -574,16 +574,16 @@ if ( navigator.mozGetUserMedia
 
     createIceServer = function (url, username, password) {
       console.warn('createIceServer is deprecated. It should be replaced with an application level implementation.');
-      
+      // Note: Google's import of AJS will auto-reverse to 'url': '...' for FF < 38
       var iceServer = null;
       var urlParts = url.split(':');
       if (urlParts[0].indexOf('stun') === 0) {
         iceServer = { urls : [url] };
       } else if (urlParts[0].indexOf('turn') === 0) {
         if (webrtcDetectedVersion < 27) {
-          var turn_urlParts = url.split('?');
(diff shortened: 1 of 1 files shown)


johache:
@letchoo that is for FF , 27 only. Was probably deprecated in their webrtc-adapter (FF 27 is pretty old).
I'm not the one who wrote this code in the first place, I would prefer to keep it as it is.

letchoo:
@johache Ok noted. Thanks. Makes sense.

johache:
Regarding FF < 38: this is managed by webrtc-adapter, we don't need to take care of it.
https://github.com/webrtc/adapter/blob/master/src/js/firefox/firefox_shim.js#L62

letchoo:
@johache ok noted. Seems fine for now then

source/adapter.js
```diff
@@ -574,25 +574,25 @@ if ( navigator.mozGetUserMedia
...
         if (webrtcDetectedVersion < 27) {
-          var turn_url_parts = url.split('?');
-          if (turn_url_parts.length === 1 ||
-            turn_url_parts[1].indexOf('transport=udp') === 0) {
+          var turnUrlParts = url.split('?');
+          if (turnUrlParts.length === 1 ||
+            turnUrlParts[1].indexOf('transport=udp') === 0) {
             iceServer = {
-              url : turn_url_parts[0],
+              urls : [turn_urlParts[0]],
```
johache: @letchoo Just saw that there was an error here, will be sending a new PR to correct that in a minute

PR #160, opened 2016-02-17
AJS-243 update googles adapterjs submodule

johache:
* updated webrtc-adapter submodule
* greatly improves Egde support
* added code in preparation of when webrtc-adapter is compiled using grunt

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -19,7 +19,7 @@ module.exports = function(grunt) {
 
       source: 'source',
 
-      googleAdapterPath: 'third_party/adapter/adapter.js',
+      googleAdapterPath: 'third_party/adapter/out/adapter.js',
 
       production: 'publish',
 
@@ -303,6 +303,19 @@ module.exports = function(grunt) {
       }
     });
 
+    grunt.registerTask('webrtc-adapter', 'Build the webrtc-adapter submodule', function() {
+      grunt.verbose.writeln('Spawning child process to compile webrtc-adapter subgrunt.');
+      var done = this.async();
+      var child = grunt.util.spawn({
+        grunt: true,
+        args: ['--gruntfile', './third_party/adapter/Gruntfile.js', 'build'],
+        opts: {stdio: 'inherit'},
+      }, function(error, result) {});
+      child.on('close', function (code) {
+        done(code === 0);
+      });
+    });
+
     grunt.registerTask('dev', [
         'CheckPluginInfo',
         'versionise',
@@ -315,6 +328,7 @@ module.exports = function(grunt) {
 
     grunt.registerTask('publish', [
         'CheckPluginInfo',
+        'webrtc-adapter',
         'versionise',
         'clean:production',
         'concat',
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -19,7 +19,7 @@ module.exports = function(grunt) {
 
       source: 'source',
 
-      googleAdapterPath: 'third_party/adapter/out/adapter.js',
+      googleAdapterPath: 'third_party/adapter/adapter.js',
 
       production: 'publish',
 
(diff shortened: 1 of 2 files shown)


letchoo:
The upgrading code seems fine.

Just curious that I see that the commits has one stating that webrtc-adapter is downgraded to 0.2.9 when the latest version is 0.2.9 ([See releases](https://github.com/webrtc/adapter/releases)). Is it that the latest master code changes is different from the one in 0.2.9 release?

johache:
it is, very different.
They seem to be moving towards an adapter compiled from several files using grunt.
This PR includes a grunt task that will build it.

They also seem to be moving towards an npm/bower kind of use. They will probably start depending on require soon.

letchoo:
Ok it's a good thing we don't move. Not sure how are we going to handle from here after they move to require format.

I'll start testing now

johache:
We want AJS to be a drop-in replacement of webrtc-adapter. 
If they are moving away from a DOM link into a require dependency, I suppose we will do the same and bump one major version up.

We can also probably generate several versions of AJS, some with require, some without.

letchoo:
## Test cases
Demos based off [Google-WebRTC-Samples repo](https://github.com/Temasys/Google-WebRTC-Samples). Branch name `jo-local-adapter`. Following demos according to latest commits which index is replaced to point to local version of adapterjs.

#### Browser: Safari _Version 8.0.6_
- (PASS) `peerconnection/recall/`
- (PASS) `peerconnection/multiple-relay/`
- (PASS) `peerconnection/multiple/`
- (FAIL) `peerconnection/dtmf` | Gets an alert of "This demo requires the RTCPeerConnection method createDTMFSender() which is not support by this browser.".
- (PASS) `peerconnection/add-stream-dynamically/`
- (WARN) `peerconnection/munge-sdp/` | The normal steps works fine. But if you try "Get Media" -> "Create Peer Connection" -> "Create offer" -> "Set offer" -> "Create answer" -> "Set answer" -> "Get Media" (another camera - change video/audio) -> "Create offer" -> "Set offer" -> "Create answer" -> "Set answer" -> Empty black stream.
- (PASS) `peerconnection/multiple-streams/`
- (PASS) `peerconnection/audio/`
- (PASS) `peerconnection/pc1/`
- (FAIL) `peerconnection/pr-answer/` | Gets an empty video stream. Try "Call" -> "Accept" -> "Hang up"
- (PASS) `peerconnection/transfer-tracks/`
- (WARN) `peerconnection/trickle-ice/` | Gathering of ICE candidates is ok with removal of servers as well. It's noted that the ICE transports feature does not seem to work because it's likely not yet implemented. Also additionally, I've added my own `turn:numb.viagene.ca` server (a free service), and it seems like the plugin throws an error after adding and attempting to gather candidates. 
```
[Error] TypeError: undefined is not a function (evaluating 'AdapterJS.WebRTCPlugin.plugin.PeerConnection(AdapterJS.WebRTCPlugin.pageId,iceServers,mandatory,optional)')
	RTCPeerConnection (adapter.js, line 2)
	start (main.js, line 122)"
```
- (PASS) `datachannel/` | Both SCTP and RTP channels works
- (PASS) `datachannel-arraybuffer/` | Both SCTP and RTP channels works
- (PASS) `getusermedia/resolution/`
- (PASS) `getusermedia/source/`
- (FAIL) `getusermedia/playPause/` | Expected it to run but it does not do anything.
- (PASS) `getusermedia/gum/`
- (PASS) `getusermedia/audio-video/` | I can't hear audio but it must be my Microphone


----
I'm not sure what does `peerconnection/multiple-relay/` demo does, but I consider it a (PASS) when calling works and video does not freeze pressing the "Insert relay" button many times.

For `getusermedia/resolution/` demo, it seems like the object stays the same size. I modified the css a bit so it's easier to see if the resolution is accurate:
```
video,
  object {
    background: none;
    height: 720px;
    width: 1280px;
  }
```

For `datachannel-arraybuffer/` and `datachannel/` demo, it seems like the link to `adapter.js` file script is broken. It should be `<script src="../../js/adapter.js"></script>`. I've added `AdapterJS.webRTCReady` checks also for the "Start" button:
```
function enableStartButton() {
  AdapterJS.webrtcReady(function () {
    startButton.disabled = false;
  });
}
```

letchoo:
This is a brief test. For @johache to evaluate

johache:
`dtmf` -> upgrade plugin, tested ok 
`pr-answer` -> broken test ?
`playPause` -> tested ok 

Will resolve the conflicts and merge

Issue #163, opened 2016-02-19
global objects should be exported or set to window object


zdila:
It is impossible to get `attachMediaStream` when using webpack and source

`import AdapterJS from 'adapterjs';`

Object `AdapterJS` has no reference to `attachMediaStream` or other objects/functions.

`window.RTCPeerConnection` is available (because it is set to the `window` object)

johache:
I am not familiar with webpack.
attachMediaStream is an attribute of AdapterJS. It is a global function. Did you take this in account ?

zdila:
Maybe I do something wrong, but in my case `AdapterJS.attachMediaStream` is `undefined`. It contains only following properties:
&gt; ` Object.keys(AdapterJS)`
< `["options", "VERSION", "onwebrtcready", "_onwebrtcreadies", "webRTCReady", "WebRTCPlugin", "onwebrtcreadyDone", "maybeThroughWebRTCReady", "TEXT", "_iceConnectionStates", "_iceConnectionFiredStates", "isDefined", "parseWebrtcDetectedBrowser", "maybeFixConfiguration", "addEvent", "renderNotificationBar"]`

Regarding Webpack - it just packages the modules from npm (and bower) to single `bundle.js`. If a module uses global variables, they are treaten as module-local variables and are not visible to other modules. But assigning to `window.attachMediaStream` should work. Also it is possible to tell Webpack to bind module-local variables of specific module to `window` so I can work it around. They say it is for "broken" modules.

johache:
oops, I missed the not...
attachMediaStream is *NOT* an attribute of AdapterJS.

you should be able to access it globally.

zdila:
OK, I can. But I must tell webpack to treat AdapterJS as "broken" module.

To "fix" it there may be more alternatives. One of:
- assign global variables to `window` (eg. `window.attachMediaStream = ...`)
- export globals from the module
- assign globals to AdapterJS object (AdapterJS is currently the only thing exported)

johache:
@ncurrier I'm fine with setting `window.attachMediaStream = attachMediaStream`
Any opinion on this ?

zdila:
It was hard by my workaround (using Webpack) is:

```js
const ajs = require("imports?module=>false!exports?AdapterJS&attachMediaStream&getUserMedia&RTCPeerConnection&createIceServer!adapterjs");
const {AdapterJS, attachMediaStream, getUserMedia, RTCPeerConnection, createIceServer} = {...ajs};
```

[...]

PR #165, opened 2016-02-23
AJS-245: propagated attachMediaStream and reattachMediaStream to window and AdapterJS

johache:
What the title says :)

In response of this issues : 
https://github.com/Temasys/AdapterJS/issues/163

### DIFF
--- a/source/adapter.js
+++ b/source/adapter.js
@@ -681,6 +681,12 @@ if ( navigator.mozGetUserMedia ||
     return to;
   };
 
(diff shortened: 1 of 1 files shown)


letchoo:
Looks like a good initiative. :)

Btw I'm not sure if you want to export RTCPeerConnection as window as well.

https://github.com/webrtc/adapter/blob/master/src/js/chrome/chrome_shim.js#L86

johache:
That is a good point.
Lets merge this first, we can consider exporting more in the future.

PR #166, opened 2016-02-23
BUGFIX: bad refactor in previous PR;  FF < 27

johache:
I seem to have missed a variable in my previous 'createIceServer' refactor. 
This fixes it.


letchoo:
:+1: good thing it's pointed out!

PR #170, opened 2016-03-04
Fixing screensharing-gum link in Chrome

johache:
With the screensharing AJS, Chrome currently takes a screensharing request as a camera request.
Seems to be due to a bad gum link.

This should fix it.

### DIFF
--- a/source/adapter.screenshare.js
+++ b/source/adapter.screenshare.js
@@ -149,6 +149,7 @@
     };
 
     getUserMedia = navigator.getUserMedia;
+    navigator.mediaDevices.getUserMedia = requestUserMedia;
 
   } else if (navigator.mediaDevices && navigator.userAgent.match(/Edge\/(\d+).(\d+)$/)) {
     // nothing here because edge does not support screensharing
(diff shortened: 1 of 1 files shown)


letchoo:
@johache looks fine but we have to do that for Firefox and Safari/IE as well.

johache:
I works fine for me in FF, Safari and IE.

letchoo:
@johache, I noticed that it triggers the error twice for Chrome.

What I did:
```
var p = navigator.mediaDevices.getUserMedia({ video: { mediaSource: 'window' } });
p.then(function(mediaStream) { console.info(mediaStream); })
p.catch(function(error) { throw error; })
```

It has to reference for the FF because of the screensharing permission prompt. Please revise it.

For Safari, I cannot get any errors (I not using a Safari screensharing plugin), which I should because in the code it checks if the plugin has screensharing feature.

johache:
You're right, updated.
The popup on Firefox will required this PR to be merged too: https://github.com/Temasys/AdapterJS/pull/171/files

johache:
Managed by https://github.com/Temasys/AdapterJS/pull/173

PR #174, opened 2016-03-07
Fix getUserMedia when not imported in global scope (ex. via webpack)

joshuawarner32:
When imported in global scope (say, directly in a script tag), the existing code works fine because `window.getUserMedia = ...` also assigns to the `getUserMedia` variable declared as part of the included google adapter.js code.

However, when the code isn't imported in global scope (for example, if it's included via webpack - where it's evaluated inside a function), `window.getUserMedia` and `getUserMedia` are actually two different variables.  This causes `requestUserMedia` (defined in the google adapter.js code) to fail when it tries to read the local `getUserMedia` variable, which will be null at this point.

I can add a commit to update the stuff in `publish` (calling `grunt publish` after initializing the submodule), if you want.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -293,6 +293,12 @@ module.exports = function(grunt) {
         grunt.log.writeln('bamboo/vars file successfully created');
     });
 
+    grunt.registerTask('CheckSubmodules', 'Checks that third_party/adapter is properly checked out', function() {
+      if(!grunt.file.exists(grunt.config.get('googleAdapterPath'))) {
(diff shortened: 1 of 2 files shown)


joshuawarner32:
I added a commit to fail the build with a helpful message when running `grunt publish` without the submodules checked out.  That should help avoid the confusion I ran into in the future (where the output mysteriously didn't have the google adapter code).

That commit is semantically unrelated to the original PR, and I'm happy to back that out or move it to another PR, at your preference.

johache:
Hi,
The content looks good to me (although I am not sure the referencing on Edge is necessary).

I'll merge and do some minor modification on the grunt task error.
Thanks for the work :)

PR #175, opened 2016-03-08
TWP-328 Forward all properties of video element to the plugin object

johache:
Rebased and corrected Seb's branch
- forward attributes to the plugin
- concerns only the plugin (IE/Safari), not Chrome and FF

How to test:
* have a <video> element
* set it with a few attributes
* call gUM and `video = attachMediaStream(video, stream)`
* test that the new video element has the same attributes as the original one

### DIFF
--- a/source/adapter.js
+++ b/source/adapter.js
@@ -1118,6 +1118,12 @@ if ( navigator.mozGetUserMedia ||
     AdapterJS.reattachMediaStream = reattachMediaStream;
 
     AdapterJS.forwardEventHandlers = function (destElem, srcElem, prototype) {
+      for (var property in srcElem) {
+        if (srcElem.hasOwnProperty(property)) {
+          destElem[property] = srcElem[property];
+        }
+      }
+
       properties = Object.getOwnPropertyNames( prototype );
       for(var prop in properties) {
         if (prop) {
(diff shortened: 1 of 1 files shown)


letchoo:
`muted` and `autoplay` does not get reflected in the new element.

johache:
The following fixes the issue on Safari, but breaks IE. 
I think think this is needed for this release, I suggest we get back to it later

```
    AdapterJS.forwardAttributesAndEvents = function (destElem, srcElem, prototype) {
      for (var property in srcElem) {
        if (srcElem.hasOwnProperty(property)) {
          destElem[property] = srcElem[property];
        }
      }

      properties = Object.getOwnPropertyNames( prototype );
      for(var prop in properties) {
        if (prop) {
          propName = properties[prop];

          if (typeof(propName.slice) === 'function') {
            if (propName.slice(0,2) === 'on' && srcElem[propName] !== null) {
              AdapterJS.addEvent(destElem, propName.slice(2), srcElem[propName]);
            } else {
              try {
                destElem[propName] = srcElem[propName];
              } catch(e) {
                // Some properties are readonly and will throw errors when set
                continue;
              }
            }
          }
        }
      }
      var subPrototype = Object.getPrototypeOf(prototype);
      if(!!subPrototype) {
        AdapterJS.forwardAttributesAndEvents(destElem, srcElem, subPrototype);
      }
    };

```

Issue #177, opened 2016-03-12
RevokeMediaStream


hoene:
Using Chrome Version 49.0.2623.75 unknown (64-bit) on Ubuntu 14.04 and adapter.js 0.13.0, I had some issues calling attachMediaStream(something, null) to detach a media stream.

the error message is 

angular.js:12722 TypeError: Failed to execute 'createObjectURL' on 'URL': No function was found that matched the signature provided.
    at TypeError (native)
    at HTMLVideoElement.Object.defineProperty.set (/bower_components/adapter.js/index.js:630:28)
    at attachMediaStream (/bower_components/adapter.js/index.js:1028:27)
    at attachMediaStream (/bower_components/adapter.js/index.js:1242:5)
    at updateMedia (/src/controllers/meeting-controller.js:291:29)
    at Object.video (/src/services/video-service.js:87:21)

johache:
Pull request opened : https://github.com/Temasys/AdapterJS/pull/178

johache:
Hi,
This issue should be fixed with today's release:
https://github.com/Temasys/AdapterJS/releases/tag/0.13.1

I'm closing the issue, feel free to re-open it if you still see it.

PR #178, opened 2016-03-14
github#177 special case for attachMediaStream on Chrome when no stream

johache:
In answer of https://github.com/Temasys/AdapterJS/issues/177 .

Setting `srcObject` to `null` on a video object generates an error on Chrome.
Create a special case for Chrome where we set it to an empty string instead.

### DIFF
--- a/source/adapter.js
+++ b/source/adapter.js
@@ -673,7 +673,12 @@ if ( navigator.mozGetUserMedia ||
   // to support the plugin's logic
   attachMediaStream_base = attachMediaStream;
   attachMediaStream = function (element, stream) {
-    attachMediaStream_base(element, stream);
+    if (webrtcDetectedBrowser === 'chrome' && !stream) {
+      // Chrome does not support "src = null"
+      element.src = '';
+    } else {
+      attachMediaStream_base(element, stream);
+    }
     return element;
   };
   reattachMediaStream_base = reattachMediaStream;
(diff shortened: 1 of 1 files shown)


johache:
Updated with support for Opera.

letchoo:
Btw `var video = attachMediaStream(videoDOM, null)` in Safari 8.0.6 throws an error `Error: Error calling method on NPObject.`. Used plugin version `0.8.866`. In Firefox 44.0.2 (stable), it's fine. Same with Firefox Nightly 47.0a1 (beta).

letchoo:
Test changes seems fine.

- (PASS) Chrome Stable 49
- (PASS) Chrome Canary 51
- (PASS) Firefox Stable 45
- (PASS) Firefox Nightly 47.0a1 
- (PASS) Opera Stable 35
- (PASS) Opera Beta 36
- (FAILED) Safari 8.0.6 -> Returns `Error: Error calling method on NPObject.`. Functionality works but returns an error.

johache:
Good catch. It seems like Safari was updated and no longer support undefined event handlers to be set on plugins ? Or our implementation no longer supports it.
Anyway, I added a check that the function exists.

letchoo:
Tested. It seems to work fine. On the side note, should we add checks if the video element is defined? Not importance as of now, not something we need in this PR.

PR #184, opened 2016-03-17
AJS-257: Fixes to ensure that navigator.mediaDevices is always defined

letchoo:
This should fix the issue in IE browsers that navigator.mediaDevices is not defined.

See JIRA Ticket - [AJS-257](http://jira.temasys.com.sg/browse/AJS-257)


letchoo:
This is into incorrect branch

PR #185, opened 2016-03-17
AJS-257: Fixes to ensure that navigator.mediaDevices is always defined.

letchoo:
Re-opened from #184 due to incorrect branch.

----

This should fix the issue in IE browsers that navigator.mediaDevices is not defined.

See JIRA Ticket - [AJS-257](http://jira.temasys.com.sg/browse/AJS-257)

### DIFF
--- a/source/adapter.js
+++ b/source/adapter.js
@@ -760,6 +760,11 @@ if ( navigator.mozGetUserMedia ||
   AdapterJS.parseWebrtcDetectedBrowser();
   isIE = webrtcDetectedBrowser === 'IE';
 
+  // always define navigator.mediaDevices if it's not defined
+  if (!navigator.mediaDevices) {
+    navigator.mediaDevices = {};
+  }
+
   /* jshint -W035 */
   AdapterJS.WebRTCPlugin.WaitForPluginReady = function() {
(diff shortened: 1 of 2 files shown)


letchoo:
Tested in both Safari 8.0.6 (Plugin 0.8.866) and IE 9 / 11.

source/adapter.js
```diff
@@ -999,8 +1004,7 @@ if ( navigator.mozGetUserMedia ||
     window.navigator.getUserMedia = getUserMedia;
 
     // Defined mediaDevices when promises are available
-    if ( !navigator.mediaDevices &&
-      typeof Promise !== 'undefined') {
+    if (typeof Promise !== 'undefined') {
```
letchoo: I'm not sure if my fixes will break the original purpose that you have written in here. But my fixes is to ensure that `navigator.mediaDevices` is always defined.

johache:
@letchoo how about this instead ?
Can you test and tell me if that works for you.

Reasoning is as follow:
Even if we define mediaDevices, mediaDevices.getUserMedia / requesteMedia will still be undefined.
Better not defining anything at all.
When we add promises support to IE 11 and less, that issue will be fixed.

letchoo:
Seems fine. Tested with Safari 8.0.6 and IE 9

PR #191, opened 2016-04-13
TWP-443: Fixes to ensure "relay" candidates are returned if TURN credentials are provided

letchoo:
> NOTE: Still requiring testing

This commit simply copies logic for both new and old format of constructing the plugin `RTCPeerConnection` object.

### DIFF
--- a/source/adapter.js
+++ b/source/adapter.js
@@ -959,6 +959,7 @@ if ( navigator.mozGetUserMedia ||
       if (servers && Array.isArray(servers.iceServers)) {
         iceServers = servers.iceServers;
         for (var i = 0; i < iceServers.length; i++) {
+          // Legacy plugin versions compatibility
           if (iceServers[i].urls && !iceServers[i].url) {
             iceServers[i].url = iceServers[i].urls;
           }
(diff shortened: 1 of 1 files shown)


source/adapter.js
```diff
@@ -953,24 +953,29 @@ if ( navigator.mozGetUserMedia ||
...
       // Call relevant PeerConnection constructor according to plugin version
       AdapterJS.WebRTCPlugin.WaitForPluginReady();
+
+      // RTCPeerConnection prototype from the old spec
+      var iceServers = null;
+      if (servers && Array.isArray(servers.iceServers)) {
+        iceServers = servers.iceServers;
+        for (var i = 0; i < iceServers.length; i++) {
+          if (iceServers[i].urls && !iceServers[i].url) {
+            iceServers[i].url = iceServers[i].urls;
```
letchoo: I'm not sure if the newer versions of plugin has removed `.url` support. @johache. Should we do a separate filtering for `.urls` support? Maybe not.

letchoo: Already checked with @johache. The plugin will take `.url` by default and fallback to `.urls` as backup.

letchoo:
Tested and it seems like it generates "relay" candidates. TODO: Test the GoogleWebRTC-Samples branch.

johache:
@letchoo 
Did minor modifications, please check if that's fine by you :)

source/adapter.js
```diff
@@ -971,10 +972,8 @@ if ( navigator.mozGetUserMedia ||
       if (AdapterJS.WebRTCPlugin.plugin.PEER_CONNECTION_VERSION &&
           AdapterJS.WebRTCPlugin.plugin.PEER_CONNECTION_VERSION > 1) {
         // RTCPeerConnection prototype from the new spec
-        if (iceServers) {
-          servers.iceServers = iceServers;
-        }
-        return AdapterJS.WebRTCPlugin.plugin.PeerConnection(servers);
+        var s = { iceServers: iceServers };
```
letchoo: Hmm, if I'm not sure if this is a good choice since if I pass `{ iceTransportsPolicy: 'relay', iceServers: theServers }` it will only enforce using `iceServers` in the `RTCConfiguration`.

letchoo: Perhaps we can clone the object instead and overwrite the iceServers.

johache: hum, no, I like this even less.
I'll roll back to your version

PR #203, opened 2016-05-27
Updated with more user agents

letchoo:
*Changes*
- Added `"android"` in Android and its version (float) detection.
- Added `"firefox"` in iOS and its version (int) detection.
- Added `"chrome"` in iOS and its version (int) detection.
- Added `"bowser"` in iOS and its version (float) detection.
- Added `"unknown"` in iOS with version set to `0` when it's not any of the listed browser.


letchoo:
FYI, place this on hold for now until I do more intensive tests.

letchoo:
These changes are merged with #216 PR changes.

PR #207, opened 2016-07-20
fix adapter.screenshare.js  for Edge

blackneck:
without this fix demo https://github.com/Temasys/Google-WebRTC-Samples/tree/master/src/content/peerconnection/pc1 changed to do connection for remote users with socket.io didnt work

### DIFF
--- a/publish/adapter.screenshare.js
+++ b/publish/adapter.screenshare.js
@@ -1527,7 +1527,7 @@ if ( navigator.mozGetUserMedia ||
                 url = server.urls[0];
               }
               if (url.indexOf('transport=udp') !== -1) {
-                self.iceServers.push({
+                self.iceOptions.iceServers.push({
                   username: server.username,
                   credential: server.credential,
                   urls: url


johache:
Hi, 
First, thanks for the PR. It's always (very) appreciated.

However, here are a few points:
1 - We never write any code in the publish folder as it is generated by grunt.
The modifications have to be done in the 'source' folder.

2 - The part of the code that you are looking at is actually compiled from Google's webrtc-adapter, in which they already made the same changes as you suggest. See:
https://github.com/webrtc/adapter/blob/master/src/js/edge/edge_shim.js#L97

3 - Thus the job is actually to update the webrtc-adapter submodule (which also leads to updating the build system)
There was a fair amount of changes made on their side. They now compile their adapter as well, so we have to integrate that. They also browserify CommonJS style code. 
That should not be too much of a problem, but still, some testing will need to be done.

4 - Lastly, for any future PR, please make it to the dev branch.

I think it is becoming more and more critical for us to upgrade AJS' dependencies. I hope to squeeze that in the next couple of weeks. If you want to give it a go on your side and report any issue, you are most welcome to do so.

Cheers,
J-O

letchoo:
@johache I agree with that stance.

letchoo:
Hi @blackneck, this fix that you are making seems to be from AdapterJS `0.13.3` release version [based on this line](https://github.com/Temasys/AdapterJS/blob/0.13.3/publish/adapter.screenshare.js#L1530). As stated above we do not recommend modifications to the `publish/` files and the segment of the code that you are modifying comes from the webrtc/adapter project release [`0.2.9`](https://github.com/webrtc/adapter/releases/tag/v0.2.9), which in the next release we are updated to the project release version [`2.0.3`](https://github.com/webrtc/adapter/releases/tag/v2.0.3).

PR #211, opened 2016-08-02
Allow for pluginInfo overrides

PetroccoCo:
This changes allows the AdapterJS.WebRTCPlugin.pluginInfo object to be set externally via a shim.
Making it easier for organizations that have purchased a Temasys plugin to pull the latest AdapterJS without needed to maintain custom copies.

For those of you who are interested we are setting this information via a SystemJS shim like so,
in the package.json overrides section:
```javascript
"meta": {
  "<path/to/adapter/file/inside/of/module>/adapter.js": {
    "globals": {
      "AdapterJS": "<path/to/your/shim/file>/adapterGlobalShim.js"
     },
     "format": "cjs"
   }
}
```

### DIFF
--- a/source/pluginInfo.js
+++ b/source/pluginInfo.js
@@ -1,4 +1,4 @@
-AdapterJS.WebRTCPlugin.pluginInfo = {
+AdapterJS.WebRTCPlugin.pluginInfo = AdapterJS.WebRTCPlugin.pluginInfo || {
   prefix : 'Tem',
   plugName : 'TemWebRTCPlugin',
   pluginId : 'plugin0',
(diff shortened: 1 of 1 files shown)


johache:
Hi, first of, thanks for the pull request, always appreciated.

Please note that here is already a way of building custom AdapterJS. You can use : 
grunt publish --pluginInfoRoot=pathToFolderContainingConfig --pluginInfoFile=configFileName.
The arguments respectively default to './source', and 'pluginInfo.js'

I don't know if that removed the need for this feature for you. 
If not, I have nothing against the idea, but here are a few changes that I need first: 
- please make PR to dev, not master,
- AdapterJS.WebRTCPlugin.pluginInfo.downloadLink is optional. Just fill it with AdapterJS.WebRTCPlugin.pluginInfo.downloadLinks.mac/win, or leave it undefined.

Cheers,
J-O

PetroccoCo:
Hey @johache,
My needs are for one code base to handle multiple different plugins at run time, so these changes are needed for my purposes.
I'll make the requested changes and switch the PR destination.
Thanks!

PR #212, opened 2016-08-03
Allow pluginInfo to be set at runtime

PetroccoCo:
This change allows the pluginInfo object to be modified at run time by a shim.


johache:
LGTM, thanks for the good work :)

PR #219, opened 2016-09-21
safari 10 compatibility with the plugin

PatrickJAMET:
navigator.plugin has been removed in Safari 10 and adapterJs use now navigator.mimeTypes instead to get the plugin list
QUOTE: Changed instantiation of WebKit plug-ins to happen at layout time, instead of at style resolution time
from Release 12 of the Safari Tech preview https://developer.apple.com/safari/technology-preview/release-notes/
Thus plugin won't work by default on the Google WebRTC samples


letchoo:
Approved.

johache:
Tested against Safari 9 and 10, as well as Windows with IE 11 (just in case...)
Looks good to me, will merge in dev and master, and do a general release today.

johache:
It looks a bit weird on the tree, but it seems to be merged

PR #223, opened 2016-10-03
AJS-278: Handle deprecation for "moz" interfaces correctly.

letchoo:
This fixes the issue:
> WebRTC interfaces with the "moz" prefix (mozRTCPeerConnection, mozRTCSessionDescription, mozRTCIceCandidate) have been deprecated.

Since in [`webrtc/adapter` line](https://github.com/webrtc/adapter/blob/v2.0.3/src/js/firefox/firefox_shim.js#L65) it checks if the undeprecated `RTCPeerConnection` API exists before shimming it, but [in this line](https://github.com/Temasys/AdapterJS/blob/0.14.0/source/adapter.js#L579), we define the interfaces as `null` again, hence making `webrtc/adapter` code to polyfill it although it has been deprecated.


johache:
Looks good at first glance, I'll run some quick tests later today.

johache:
LGTM.

Issue #222, opened 2016-10-03
window.propName global variable


ibc:
Here a global variable is being created:

https://github.com/Temasys/AdapterJS/blob/master/source/adapter.js#L1325

ibc:
Also `properties` in https://github.com/Temasys/AdapterJS/blob/master/source/adapter.js#L1322

johache:
yep, easy PR : https://github.com/Temasys/AdapterJS/pull/224.
thx.

PR #224, opened 2016-10-04
GH-222 : clean global variables

johache:
Removed a couple of global variables, as listed by Iñaki here : https://github.com/Temasys/AdapterJS/issues/222


letchoo:
Haven't tested but this looks legit.

PR #226, opened 2016-10-05
AJS version number with submodule hash

johache:
Show the webrtc-adapter submodule hash on AJS version number


johache:
Just realised this was never reviewed, I'm merging it to save some time.

PR #227, opened 2016-10-14
Adds forceGetAllCams if enabled in attachMediaStream.

damencho:
When enabled AdapterJS.options.getAllCams this adds the needed params when executing attachMediaStream.
It is still not working for me for safari with latest plugin, but at least has the needed param.


johache:
Hi,

You shouldn't need to do that, as long as you keep doing your gUM call on the initial plugin.
That is, AdapterJS.WebRTCPlugin.plugin, or, the <object> with id="plugin0".

PR #231, opened 2016-10-17
github#226: added support for NotAllowedError on FF 

johache:
when the screensharing extension in not available, FF now seems to return a ``NotAllowedError`` instead of ``PermissionDeniedError`` or ``SecurityError``.
Just add the former to the list of expected errors when the extension is not available.


letchoo:
Legit. +1

Issue #229, opened 2016-10-16
Use https instead of http


zdila:
Please replace links like http://skylink.io/plugin/, http://bit.ly/webrtcpluginpkg, http://bit.ly/webrtcpluginmsi with https versions.

johache:
PR merged, thx :)

PR #232, opened 2016-10-17
github#229 HTTPS links

johache:
From https://github.com/Temasys/AdapterJS/issues/229
"replace links like http://skylink.io/plugin/, http://bit.ly/webrtcpluginpkg, http://bit.ly/webrtcpluginmsi with https versions."


letchoo:
Legit. +1

PR #241, opened 2016-11-03
Remove weird chars

### LINKED ISSUE #240: Uncaught SyntaxError: Invalid or unexpected token
It seems like there are weird characters in the script when loaded in the CDN demos. See [sample demo link here](https://temasys-cdn.s3.amazonaws.com/demos/support-web/adapterjs-0.14.0-encoding.html).  If you open the Web Console, you will see the following line:
> adapter.debug.js:644 Uncaught SyntaxError: Invalid or unexpected token

Then if you click the the `adapter.debug.js` part in that console log line, you will have to scroll to line `644`, and you will see weird characters `â€` in the script. Here's an example of how it's like now:

```
// Check for browser types and react accordingly
if ( (navigator.mozGetUserMedia ||
      navigator.webkitGetUserMedia ||
      (navigator.mediaDevices &&
       navigator.userAgent.match(/Edge\/(\d+).(\d+)$/)))
    && !((navigator.userAgent.match(/android/ig) || []).length === 0 &&
      â€‚â€‚â€‚â€‚(navigator.userAgent.match(/chrome/ig) || []).length === 0 && navigator.userAgent.indexOf('Safari/') > 0)) {
```

The solution right now is to set the `charset="utf-8"` on the `<script>`, or to remove these characters in grunt tasks when compiling for SkylinkJS.

letchoo:
We remove the characters we cannot see. Fixes #240.


johache:
LGTM (although, why is the demo file removed again ?)

PR #246, opened 2016-12-05
Add AdapterJS to window object

aszmyd:
I've added this one because when i requite this plugin with webpack, it does not put the `AdapterJS` object into global `window` object but i need it to detect browser support and plugin installation status.


johache:
@aszmyd, please do you PR on dev, not master.
Also, you don't need to push  modifications on publish, we will re-generate these files on the next release.

@ncurrier I'm not sure we should put AdapterJS in the window scope. That sounds against basic npm encapsulation principles. What do you think ?

aszmyd:
@johache is it an issue that those changes are in publish folder also? Should i revert them?

Issue #69, opened 2015-04-28
Implementing plugin callback without touching AdapterJS file


ulifigueroa:
If I want to implement the`AdapterJS.WebRTCPlugin.pluginNeededButNotInstalledCb` callback the source code says that it needs to be overridden but what I don't want to do is to end up using a custom AdapterJS version. Also it would be nice to be able to pass options. For example like this:

```javascript
window.AdapterJS = {
    options: {
        hidePluginInstallPrompt: true
    },
    WebRTCPlugin: {
        pluginNeededButNotInstalledCb: function() {
            //Do something when the plugin is not installed
        }
    }
};
```

serrynaimo:
After importing AdapterJS you can just do
AdapterJS.WebRTCPlugin.pluginNeededButNotInstalledCb = function() {
    // your code
};

This way you don't need to touch the AdapterJS code.

Hope this helps :)

ulifigueroa:
I forgot to mention that I'm lazy loading AdapterJS this might be causing that at the point where I want to define the callback either AdapterJS is still not defined or the callback was already set to null because of this line https://github.com/Temasys/AdapterJS/blob/master/source/adapter.js#L122

serrynaimo:
I see. Now it makes more sense to pass it as a config. I'll keep it open for @johache to consider.

Thanks!

johache:
Hi,

Good points I think, I saw you PR : https://github.com/Temasys/AdapterJS/pull/70/files
Just looking at it it looks quite good, but I will need to run tests before validating it.
Hopefully I can get to it by Monday (hopefully...)

Thanks

RuneAndersen:
Sorry but I don't get it :-/

```AdapterJS.WebRTCPlugin.pluginNeededButNotInstalledCb = AdapterJS.WebRTCPlugin.pluginNeededButNotInstalledCb ||
      function () {...}```

I can replace the AdapterJS.WebRTCPlugin.pluginNeededButNotInstalledCb function with my own function, but it seems to me the original AdapterJS.WebRTCPlugin.pluginNeededButNotInstalledCb is called immediately after the AdapterJS script loads so my replacement has no effect and I can't figure out to it before AdapterJS loads, is that possible?.

Could you please provide me with a small working sample of how to intercept the callback method?

johache:
The callback function is indeed called immediately as AJS in loaded.
[...]

PR #253, opened 2017-01-19
get config values from global object #69

elbecita:
This is needed for AdapterJS to be able to pick values that have been set before it is required or imported, otherwise the `var AdapterJS = ...` itself is already overwriting whatever was in the global variable. 
It is interpreted as:
```
var AdaptjerJS;
AdaptjerJS = AdapterJS || {};
```
Accessing through window solves the issue.
Does it make sense to you?


johache:
LGTM.
@letchoo any objections ?

elbecita:
Actually, I think the way I did it, the real AdapterJS would be available in global scope. What do you think about getting only a copy of the config values with
`var AdapterJS = (window && JSON.parse(JSON.stringify(window.AdapterJS))) || {};` ? 

This way the real AdapterJS truly remains local and won't be available in the global scope.

elbecita:
Will it be possible to merge this? :)

PR #256, opened 2017-02-21
Update bowser support on README

letchoo:
It seems like bowser (current latest version is `0.6.1`) does not work with iOS 10+. Tested and seems like it only works on iOS 9+.

See the related ticket: [EricssonResearch/bowser#97](https://github.com/EricssonResearch/bowser/issues/97)


johache:
"iOS 9+ and bellow" ?

How about just iOS 9.x then ? ^^

letchoo:
Yeahh makes sense :)

johache:
Sorry, but I think it should just say "9.x only"
(In general, I think we should avoid emoticons in READ.MEs

letchoo:
@johache ahh ok it's not emoticons but something like footnotes (**). I guess it got confused as emoticon Lol!. Will make the update again.

PR #257, opened 2017-02-22
AJS-250: Fixes for "isIE" undefined error throwing when clicking button on notification bar

letchoo:
**What is the issue:**
image
When clicking on a button on a rendered notification bar that is on non-plugin browsers (chrome/firefox/opera), the "isIE is not defined" error throws. This is likely that "isIE" variable is only defined when case is plugin browser.

**How to reproduce it:**
- Go to a chrome/firefox/opera browser (non-plugin supported browser)
- Access any plugin demo. E.g.: https://plugin.temasys.com.sg/demo/src/content/peerconnection/trickle-ice/
- Key in this notification bar rendering to stimulate triggering for requiring add-on installation: AdapterJS.renderNotificationBar('install message', 'clickforerror', window.location.href, false, false);
- Click the "clickforerror" button
- Notice that "isIE" not defined error throws 

----
This PR aims to prevent the error from throwing and defining a variable to let `AdapterJS.renderNotificationBar` know not to run the plugin is installed interval.

### DIFF
--- a/source/adapter.js
+++ b/source/adapter.js
@@ -1,5 +1,6 @@
 // Adapter's interface.
 var AdapterJS = AdapterJS || {};
+AdapterJS.isPluginNotificationBar = false;
 
 // Browserify compatibility
 if(typeof exports !== 'undefined') {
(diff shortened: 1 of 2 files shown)


johache:
This ``renderNotificationBar`` is getting ridiculous.
Here is what I suggest : 
``AdapterJS.renderNotificationBar = function (text, buttonText, yesCallback, noCallback)``
When the user click the `okay` button, the yesCallback is triggered. Same thing for `cancel`

That would clean things up and make the notificationBar really re-usable.

letchoo:
Makes sense.

letchoo:
@johache updated PR.

johache:
The refactoring looks good.
Screensharing fails on Chrome and FF on GAR, but I guess that's because of the require issue you are working on on that other PR https://github.com/Temasys/AdapterJS/pull/258

On my basic tests, it seems to work on Chrome, but I get no popup asking me to install an extension on Firefox, I just get a skylink error "NotAllowError"

letchoo:
@johache PR #249 should resolve the "NotAllowedError" due to the error name changes for Firefox. We used to read it from "SecurityError" or "PermissionDeniedError".

johache:
Indeed.

PR #258, opened 2017-02-23
AJS-283: Temp files for require("./utils") not found and requirejs async calls in webrtc/adapterjs browserified dependency

### LINKED ISSUE #238: Error: module "./utils" not found from "***adapter.min.js"
Hi!
   I couldn't found 'utils' anywhere.

letchoo:
This should fix #238 for now.

Currently while we look into make AdapterJS handle the browserify issues with webrtc/adapter dependency when we have the time, this temp fixes if it doesn't cause any issues should resolve the issue:

- Replace the `require(` to `requirecopy(` in the browserified file.
- Since `AdapterJS.webRTCReady()` is required initialization method to be invoked, define the screensharing overrides if requirejs module is defined.

Tested on:
- SkylinkJS tape tests which has the `require("./utils")` not found error and it works.
- SkylinkJS demo that doesn't use requirejs and it works.
- SkylinkJS getaroom.io demo and it works, the screensharing polyfills were called.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -11,6 +11,7 @@ module.exports = function(grunt) {
     grunt.loadNpmTasks('grunt-include-replace');
     grunt.loadNpmTasks('grunt-karma');
     grunt.loadNpmTasks('grunt-githash');
+    grunt.loadNpmTasks('grunt-string-replace');
 
     grunt.initConfig({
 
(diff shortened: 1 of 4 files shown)


johache:
1. ``AdapterJS.defineMediaSourcePolyfill`` is undefined but called in ``adapter.debug.js``
``AdapterJS.webrtcReady`` is **optionally** called by the user, it is not a good place for that call. 
Maybe consider ``AdapterJS.maybeThroughWebRTCReady `` (I'm not saying that's the right place).
2. On Safari, with the plugin un-installed, you will be calling ``AdapterJS.defineMediaSourcePolyfill`` before getUserMedia is set, so baseGetUserMedia is undefined, which will definitely lead to problems.

letchoo:
Erm actually we tell users to require it as an requirement -  https://github.com/Temasys/AdapterJS#required-initialization. I remembered we need it for Safari / IE plugins or something previously, and `AdapterJS.webRTCReady()` would do the job of handling both cases or something.

letchoo:
@johache wait let me force-push to remove the publish changes.

johache:
LGTM

PR #263, opened 2017-04-06
TWP-681 Added MediaStream constructor

miniruwan:
Added MediaStream constructor as requested in https://jira.temasys.com.sg/browse/TWP-681

### DIFF
--- a/source/adapter.js
+++ b/source/adapter.js
@@ -561,6 +561,9 @@ createIceServer = null;
 createIceServers = null;
 //------------------------------------------------------------
 
+//Creates MediaStream object.
+MediaStream = null;
+
 //The RTCPeerConnection object.
 RTCPeerConnection = (typeof RTCPeerConnection === 'function') ?
   RTCPeerConnection : null;
@@ -833,6 +836,7 @@ if ( (navigator.mozGetUserMedia ||
   //
   // Shims the follwing:
   // -- getUserMedia
+  // -- MediaStream
   // -- MediaStreamTrack
   // -- MediaStreamTrack.getSources
   // -- RTCPeerConnection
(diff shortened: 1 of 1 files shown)


source/adapter.js
```diff
@@ -561,6 +561,9 @@ createIceServer = null;
 createIceServers = null;
 //------------------------------------------------------------
 
+//Creates MediaStream object.
+MediaStream = null;
```
letchoo: @miniruwan I think we should not set the `MediaStream` to `null` because this could be defined for other browsers.

letchoo:
Seems ok for me now.

source/adapter.js
```diff
@@ -1036,6 +1040,14 @@ if ( (navigator.mozGetUserMedia ||
         ConstructSessionDescription(info.type, info.sdp);
     };
 
+    MediaStream = function (mediaStreamOrTracks) {
+      AdapterJS.WebRTCPlugin.WaitForPluginReady();
+      if (typeof mediaStreamOrTracks !== 'undefined' && mediaStreamOrTracks !== null) {
```
johache: What happens in the plugin if you receive undefined or null (might be different on IE and Safari) ?
Could we remove that if statement ?

miniruwan: @johache yes, we can remove this if statement in javascript.
**Explanation :** In the plugin c++ code, there is an if condition which is true when mediaStreamOrTracks is not available. That if condition is true when mediaStreamOrTracks javascript object is either null or undefined. In that case, a new MediaStream object will be created without any streams in it (as if it were created without any arguments for MediaStream() constructor).

miniruwan: Done. pushed.

johache:
LGTM

PR #264, opened 2017-04-11
AJS-284 : Enable using mediaSource for screensharing with the plugin.

johache:
These modifications allow to set a specific key as a mediaSource to filter for screen only, window only, or both.

These behaviour requires a plugin v0.8.891 or higher (```AdapterJS.WebRTCPlugin.plugin.screensharingKeys```) defined.
Otherwise, it will default to the current behaviour.


letchoo:
@johache apologies to just see this. This PR related logic might have been released in `0.14.2`.

johache:
@let yep, this is no longer needed.

PR #265, opened 2017-04-17
Use native RTCPeerConnection API when available for Edge browsers on Windows Creator Update

letchoo:
- The latest version of `webrtc/adapter` uses the ORTC polyfill still, which doesn't work on the latest Edge browsers on Windows Creator Update. Since RTCPeerConnection API is available, there's no need for polyfills.
- This should prevent any overrides from the `webrtc/adapter`.

### DIFF
--- a/source/adapter.js
+++ b/source/adapter.js
@@ -612,6 +612,11 @@ if ( (navigator.mozGetUserMedia ||
   ///////////////////////////////////////////////////////////////////
   // INJECTION OF GOOGLE'S ADAPTER.JS CONTENT
 
+  // Prevent Edge overrides
+  if (navigator.userAgent.match(/Edge\/(\d+).(\d+)$/) && window.RTCPeerConnection) {
+    window.cachedRTCPeerConnection = window.RTCPeerConnection;
+  }
+
 /* jshint ignore:start */
 @Goo@include('third_party/adapter/out/adapter.js', {})
 /* jshint ignore:end */
@@ -771,6 +776,10 @@ if ( (navigator.mozGetUserMedia ||
       to.srcObject = from.srcObject;
       return to;
     };
+
+    if (window.cachedRTCPeerConnection) {
+      window.RTCPeerConnection = window.cachedRTCPeerConnection;
+    }
   }
 
   // Need to override attachMediaStream and reattachMediaStream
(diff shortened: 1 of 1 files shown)


johache:
Hum, that does't sound good to me.

If Google is shimming RTCPC, I'm guessing they have a reason.
Last time I tried, I remember the connection working Egde to Egde. Is that no longer the case ?
What is the test scenario/link ?

Also, the adapter.js repo seems to be pretty active regarding Edge support : https://github.com/webrtc/adapter/commits/master
Did you check if updating our dependency would fix you issue ?

letchoo:
@johache actually I tested the later version of their release: https://github.com/webrtc/adapter/tree/v3.3.2 and it seems like their ORTC polyfill for WebRTC API didn't work well with the Edge 15.xxxx Windows Creator Update version. 

They seem to be overriding the RTCPeerConnection API instead of using the native implemented one which is better as according to [here](https://github.com/webrtc/adapter/blob/master/src/js/edge/rtcpeerconnection_shim.js#L74).

letchoo:
@johache actually, let me just do a PR into their repo instead. Depending on the time constraints as well if we need to get our AdapterJS to use Edge native RTCPeerConnection, I'll leave this PR open for now.

fippo:
if you want Edge's native RTCPeerConnection use webkitRTCPeerConnection (which says a lot about what that implementation does) which will be left untouched by adapter.

@letchoo what seems to be broken? Still works for me (even though getUserMedia is busted in the insider builds)

letchoo:
@fippo it seems like there was some "InvalidAccessError" whilst testing with the latest master branch which had issues with the 15.xxx version and had no issues with the 14.xxx version. The native legacy RTCPeerConnection API seem to work with 15.xxx. Will probably open the relevant tickets in any case.

fippo:
@letchoo looks like there is some issue in 15063 which wasn't there in 14393. Will take a look, thanks!

johache:
@letchoo keeping msRTCPC Sounds fair, although it does not solve our actual issue. I'm happy with merging that first if you want.

@fippo We've mostly tested against our AdapterJS (built from different version of webrtc/adapter).  I feel like the best way of moving this forward would be for us to produce a minimal test case to reproduce. I'll try to do that next week.

letchoo:
Hey there @fippo, damn me but the tests for the latest webrtc was not an accurate test since it was basing off an 2.x.x version of the webrtc/adapter after I double-checked again. I was testing with the latest webrtc/adapter and it seems like there are separate issues (I think I opened one [https://github.com/webrtc/adapter/issues/510](#510)) with the connections. Thanks a lot for the investigations!

letchoo:
@johache so now it's  cached to `msRTCPeerConnection`. Is this ok to go?

johache:
Looks fine to me.

PR #270, opened 2017-05-24
AJS-284 prevent filtering of mediaSource in contraintsToPlugin

johache:
Everything is in the title...
This change doesn't actually have any impact at the moment as we still set optional[n].sourceId, but eventually, we want to move away from this system.

### DIFF
--- a/source/adapter.js
+++ b/source/adapter.js
@@ -1121,7 +1121,11 @@ if ( (navigator.mozGetUserMedia ||
       }
       var cc = {};
       Object.keys(c).forEach(function(key) {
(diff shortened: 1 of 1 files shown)


letchoo:
Tested and checked, seems fine to me. On a side note, for weird reasons, although `"mediaSource"` is supported by Firefox 53 in `navigator.mediaDevices.getUserMedia()` it's filtered away [in here](https://github.com/webrtc/adapter/blob/master/src/js/firefox/getusermedia.js#L46).

johache:
Good question.
Anyway, I'll merge.

PR #271, opened 2017-05-24
Updated README.md docs for mediaSource and configuration.

letchoo:
The title says it all. Just updated clearer docs for mediaSource available values and configuring AdpaterJS. Not to be confused with AJS-291 which will be more detailed docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,7 +257,7 @@ window.navigator.getUserMedia({
 });
 ```
 
-**List of available sources:**
+**List of valid screensharing (`mediaSource`) sources:**
 
 | Sources | Description | Browsers that supports it |
 | ------- | ----------- | --------- |
--- a/README.md
+++ b/README.md
@@ -278,88 +278,36 @@ window.navigator.getUserMedia({
 | `AdapterJS.WebRTCPlugin.plugin.screensharingKeys.window` | Deprecated. Same as `"window"`. | IE, Safari |
 
 
-## Configuring AdapterJS
+**Configuring extension settings:**
 
-To configure your AdapterJS, configure this before referencing the script for `adapter.screenshare.js` or `adapter.debug.js` as an example:
+To configure your AdapterJS screensharing extensions, configure this before referencing the script for `adapter.screenshare.js` as an example:
 
 ```
 var AdapterJS = {};
-AdapterJS.WebRTCPlugin = {};
-
-/**
- * Configure plugin info. Reference "source/pluginInfo.js"
- */
-AdapterJS.WebRTCPlugin.pluginInfo = {};
-
-/**
- * Configure plugin settings
- */
-// To get virtual webcams. false by default
-// AdapterJS.options.getAllCams = true;
-
-// To prevent the install prompt when the plugin in not yet installed. false by default
-// AdapterJS.options.hidePluginInstallPrompt = true;
-
-/**
- * Configure plugin installation bar text. Uses default when not provided.
- */
-// Configure plugin installation bar text when requiring installation
-// AdapterJS.TEXT.PLUGIN.REQUIRE_INSTALLATION = 'xxx';
-
-// Configure plugin installation bar text when not supported for WebRTC plugin and no WebRTC functionalities
-// AdapterJS.TEXT.PLUGIN.NOT_SUPPORTED = 'xxx';
-
-// Configure plugin button text.
-// AdapterJS.TEXT.PLUGIN.BUTTON = 'xxx';
-
-// Configure refresh page after plugin installation bar text
-// AdapterJS.TEXT.REFRESH.REQUIRE_REFRESH = 'xxx';
-
-// Configure refresh page after plugin installation bar button text
-// AdapterJS.TEXT.REFRESH.BUTTON = 'xxx';
-
 
 /**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -257,6 +257,111 @@ window.navigator.getUserMedia({
...
+| ------- | ----------- | --------- |
+| `"window"`  | Fetches the list of application windows | Chrome, Opera, Firefox, IE, Safari |
+| `"screen"`  | Fetches the list of display screens     | Chrome, Opera, Firefox, IE, Safari |
+| `"tab"`  | Fetches the list of browser tabs | Chrome, Opera |
+| `"browser"`  | Fetches the list of browser windows | Firefox (Requires to configure `about:config` to enable `media.getusermedia.browser.enabled`)  |
+| `"application"`  | Fetches the list of applications | Firefox |
+| `"camera"`  | Fetches the list of cameras | Firefox |
+| `["tab", "audio"]`  | Fetches `"tab"` with its audio | Chrome, Opera (Enable `constraints.audio` to retrieve tab audio) |
+| `["window", "screen"]`  | Fetches `"screen"` and `"window"` | Chrome, Opera, IE, Safari |
+| `["window", "screen", "tab"]` | Fetches `"screen"`, `"window"` and `"tab"` | Chrome, Opera, IE, Safari |
```
miniruwan: Is there a feature for "tab" in the temasys plugin for IE and safari? I was not aware.

letchoo: Thx for checking.

README.md
```diff
@@ -257,6 +257,111 @@ window.navigator.getUserMedia({
...
+| `AdapterJS.WebRTCPlugin.plugin.screensharingKeys.screen`  | Same as `"screen"` | IE, Safari |
+| `AdapterJS.WebRTCPlugin.plugin.screensharingKeys.window` | Deprecated. Same as `"window"`. | IE, Safari |
+
+
+## Configuring AdapterJS
+
+To configure your AdapterJS, configure this before referencing the script for `adapter.screenshare.js` or `adapter.debug.js` as an example:
+
+```
+var AdapterJS = {};
```
johache: I don't think those 7 lines are relevant (unless you wanna define stuff like ```AdapterJS.extensionInfo.chrome.extensionId```, in which case, yes, you obviously need to define ```AdapterJS```.
I feel that putting that in the doc is confusing.

README.md
```diff
@@ -257,6 +257,111 @@ window.navigator.getUserMedia({
 });
 ```
 
+**List of available sources:**
```
johache: I think this is in the screensharing section, but could you change that to "List of available screensharing sources" or "List of valid screensharing sources"

README.md
letchoo: I'm sure perhaps it might differ for plugin configuration but was thinking from an end-user pov, I think it might be easier to have something easier to understand and configure. Hence I left it that way and for configuring things.

```
<script>/* Modify */</script>
<script src="adapter.screenshare.js"></script>
```

README.md
letchoo: Can. Might perhaps add screensharing (`mediaSource`).

README.md
letchoo: I suggest we can rephrase the line in a way it's easier to understand, I do like having something easy to configure and understand.

README.md
letchoo: Done.

README.md
letchoo: Done.

README.md
johache: Well, in this case, you really don't need to define ```AdapterJS.WebRTCPlugin``` and ```AdapterJS.WebRTCPlugin.pluginInfo```
What I am afraid of is people defining those after requesting/injecting AJS, and overriding values by mistake.

letchoo: @johache I understand the case, but there will be these developers and some will just not read documentation. But if that's the direction we are heading for, we have to modify the way `AdapterJS.extensionInfo` is overridden and instead of invoking the _defineScreensharingPolyfill after the end of the line, but rather do it in `webRTCReady()`. So users need not define `AdapterJS` (or for docs sake the `AdapterJS.WebRTCPlugin`) but then just override and invoke `webRTCReady()`.

letchoo: Perhaps we can document it this way to be clearer:

```
<script>
// the docs here
</script>
<script src="adapter.screenshare.js"></script>
```

johache: I'm fine with that.
I guess what I am trying to say is that **no-one** should be overriding/pre-defining ```AdapterJS.WebRTCPlugin```, or any of it's attributes, including ```AdapterJS.WebRTCPlugin.pluginInfo```.
I don't see a case where anyone would need to override those; Tell me if you have one.

You can keep all the rest

letchoo: There is a case where (and it works) to configure differing plugins for testing. I just stated my point was that it might be simpler to just configure everything at a shot. That's all. If unless there has been noted issues where doing these "overridings" or "pre-defining" has caused issues or there is a logical reason for that, I don't see the problem.

letchoo: Done. Removed and kept extension settings overridable. So `AdapterJS.options` and `AdapterJS.TEXT` is something we need to actually start allowing overriding in the future. We can JIRA ticket that.

PR #275, opened 2017-06-20
TWP-721 Allow document.readystate interative

johache:
In very slow networks, waiting for ```document.readyState``` to move to ```'complete'``` might take a long time (seconds, minutes).
When the readystate is ```'interactive'```, the DOM is ready and the plugin should be able to start.


letchoo:
Lgtm.

Issue #262, opened 2017-04-05
Transitive dependencies not listed when AdapterJS used as direct dependency


NTARelix:
When `adapterjs` is a dependency of package `mypackage` and you attempt to view the dependencies of `mypackage`, none of the transitive dependencies installed by `adapterjs` show up in the dependency tree given by `npm ls` with respect to `mypackage`.

To demonstrate

1. Create new package: `$ mkdir mypackage && cd mypackage && npm init -y`
2. Install `adapterjs` as dependency: `$ npm i -S adapterjs`
3. List all dependencies installed for `mypackage`: `$ npm ls`
    ```
    mypackage@1.0.0 /Users/ntarelix/mypackage
    └── adapterjs@0.14.1
    ```
4. List all dependencies installed for `adapterjs`: `$ cd node_modules/adapterjs/third_party/adapter/ && npm ls`
    ```
    webrtc-adapter@2.0.3 /Users/ntarelix/mypackage/node_modules/adapterjs/third_party/adapter
    ├─┬ chromedriver@2.28.0
    │ ├── adm-zip@0.4.7
    │ ├── kew@0.7.0
    │ ├─┬ mkdirp@0.5.1
    │ │ └── minimist@0.0.8
    │ └── rimraf@2.6.1
    ├── eslint-config-webrtc@1.0.0
    ├─┬ faucet@0.0.1
    │ ├── defined@0.0.0
    ...
    ```

Notice how a lot more than just `adapterjs` was installed, but npm was unable to pick up on these dependencies because they were not listed as dependencies of `adapterjs`. It happens because `adapterjs`'s `package.json` has a `postinstall` script that installs a submodule, `webrtc-adapter`.

This is problematic for a number of reasons. My particular problem comes from the setup of a private npm registry where I'm publishing all of my project's dependencies on a Nexus server semi-manually. I use `npm ls` to list all my dependencies and use that list to manually publish each listed version of each package individually. If they aren't listed with `npm ls` then they don't get published and cannot be installed by other packages later. This means `adapterjs` can't be installed using exclusively my private registry unless I `npm ls` from `webrtc-adapter` and also publish each of those dependencies.

Some other problems that I believe are also caused by this:

- #225
- #234

To solve this I believe it would be best to remove the `postinstall` script so nothing additional gets installed and because the distributables are already included when I `npm i adapterjs`. Not sure what the best practice alternative would be, but maybe putting it in a `dev` script and adding 1 more step to dev setup would be good enough.

[...]

PR #282, opened 2017-08-10
webrtc-adapter dependency changed to npm dependency from git submodule

NTARelix:
AdapterJS depends on webrtc/adapter (`webrtc-adapter` on npm). We currently declare this dependency via git submodules when ideally it would be in `package.json`. There are multiple reasons for this (one of which I've run into with bug #262, as well as being unable to install this dependency in a minimal CI slave that doesn't have git).

The build involves hooking into the npm lifecycle (see `postinstall` in `package.json`) to build this `webrtc-adapter` dependency before injecting it into this package's output. I've removed the submodule in favor of a pre-built npm dependency.


johache:
I think that this is very good idea.
We never merge anything straight into master though. Could you change your PR so that it goes into dev instead ?

Thanks,
J-O H

PR #283, opened 2017-08-11
webrtc-adapter dependency to npm dependency from git submodule

NTARelix:
AdapterJS depends on webrtc/adapter (webrtc-adapter on npm). We currently declare this dependency via git submodules when ideally it would be in package.json. There are multiple reasons for this (one of which I've run into with bug #262, as well as being unable to install this dependency in a minimal CI slave that doesn't have git).

The build involves hooking into the npm lifecycle (see postinstall in package.json) to build this webrtc-adapter dependency before injecting it into this package's output. I've removed the submodule in favor of a pre-built npm dependency.


NTARelix:
I opened this pull request yesterday, but did a hard reset on master and force pushed so the old pull request got closed automatically.
https://github.com/Temasys/AdapterJS/pull/282

These changes are on the development branch now.

johache:
@letchoo LGTM.
I'll merge if that's good for you.

Gruntfile.js
```diff
@@ -23,22 +22,10 @@ module.exports = function(grunt) {
...
-      githash: {
-        submodule: {
-          options: {
-            dir: '<%= webrtcAdapterSubmodule %>'
-          }
-        },
-      },
-
-      version: '<%= pkg.version %>-<%= githash.submodule.short %>',
+      version: '<%= pkg.version %>',
```
letchoo: @johache should we keep this as discussed previously @ncurrier?

johache: Well, we would no longer be depending on a git-hash, but rather a node dep' version...
If anything, then what we should put next to our version number would be the webrtc-adapter version number.

I'm not sure if that's worth it though...

letchoo: I think that just having our version number is sufficient since we can document the webrtc/adapter version in the release notes for every release and its in the `package.json` file. 

But since this might change our release tagging process a bit, it might be better we reconfirm with @ncurrier on this.

ncurrier: I'm ok with the version being specified in the package.json as long as it explicitly specified and does not rely on ranges. This PR specifies any minor release greater than 2.0.3 (^2.0.3), that I am not ok with. 

@johache are you comfortable with only sticking to published releases? That does limit us slightly from previously where its commit based, however that may not be an issue and the simplicity may be favored.

johache: I'm fine with this limitation. We can always revert in the future if we realise this is completely blocking for us.
It does sound safer to only use released version anyway. It means that it passed Google's safety checks.

@NTARelix , could you change your PR so that it uses a specific version rather than a compatible one ? (2.0.3 instead of ^2.0.3)

PR #286, opened 2017-08-30
Safari 11 macOS and iOS intergation

johache:
Safari 11 expected to land out of beta in Sept (both macOS and iOS).
We should release our changes asap, so that users have a chance to update their site before the macOS/iOS official release.

Content of the PR: 
- update of the webrtc-adapter submodule 
- added the flag ```AdapterJS.options.forceSafariPlugin ``` to force the use of the Temasys plugin on Safari 11+
- Moved the following variables into the AdapterJS scope: `webrtcDetectedBrowser`, `webrtcDetectedVersion`, `webrtcMinimumVersion`, `webrtcDetectedType`, `webrtcDetectedDCSupport`. We should deprecate their usage as global variables.
- AJS now relying on those to identify the browser being used

Safari 11 supports : 
- gUM
- PeerConnection
- DataChannel

Known limitations of Safari Tech Preview and iOS 11
- No audiooutput in getSources/enumerateDevices
- No Screensharing support
- No DTMFSender support
- No RTCRecording support
- **Connecting PeerConnections requires it to have a non empty stream (so gUM required)
signalingState doesn't go to "close":** 
https://bugs.webkit.org/show_bug.cgi?id=174314
This creates a weird situation where, if you want to use a DC on PC, you still have to gUM and attach the stream to the PC...
- No VP8 / VP9 support. H264 only.
Means the free plugin cannot connect to Safari 11+
https://bugs.webkit.org/show_bug.cgi?id=173141


letchoo:
I'm good with that but should we test with webrtc/adapter [`5.0.1`](https://github.com/webrtc/adapter/releases/tag/v5.0.1) and consider bumping to that version? Seems like [`5.0.0`](https://github.com/webrtc/adapter/releases/tag/v5.0.0) has some MediaStream IDs mapping fixes.

johache:
@letchoo indeed... They released those yesterday...

letchoo:
Actually are we using versioning instead of submodule?

johache:
Yes, indeed.
I removed the submodule and replaced it with a node dep.

PR #287, opened 2017-09-14
Add polyfill for Array.prototype.forEach.

letchoo:
Simple PR. Noticed that `Array.prototype.forEach` is used in [line `178`](https://github.com/Temasys/AdapterJS/blob/0.15.0/source/adapter.js#L178), [line `1165`](https://github.com/Temasys/AdapterJS/blob/0.15.0/source/adapter.js#L1165) and [line `1210`](https://github.com/Temasys/AdapterJS/blob/0.15.0/source/adapter.js#L1210). But according the [MDN docs for it](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) IE 8 does not support that function natively although we don't recommend that browser version. This is just to polyfill it when it's missing to prevent errors.

### DIFF
--- a/source/adapter.js
+++ b/source/adapter.js
@@ -1,3 +1,14 @@
+// Polyfill forEach which is not supported in IE 8
+if (typeof Array.prototype.forEach !== 'function') {
+  Array.prototype.forEach = function (fn) {
+    var index = 0;
+    while (index < this.length) {
+      fn(this[index], index);
+      index++;
+    }
+  };
+}
+
 // Adapter's interface.
 var AdapterJS = AdapterJS || {};
 


kekkokk:
tried today the master branch but without polyfill I cannot access to navigator.mediaDevices.desiredfunction and other functions of the native adapter.
This way I'll have to change all my code to support AdapterJS calls instead of native ones. why don't just include a promise polyfill?

letchoo:
@johache yes that is the logic intended.

letchoo:
@johache understand. If I recall though, it works fine in my Windows XP IE 8 before. Perhaps instead of using `forEach()`, we use while or for loops instead so we don't limit completely the support?

johache:
As a default, I'm against making changes to support browsers that we don't advertise support for...
Just because it happened to work at some point (by luck) doesn't mean we should go and alter native prototypes to maintain support in the present.

@ncurrier to arbitrate here

letchoo:
I understood your point and stance, but I'm asking that instead of using `forEach()`, use `for` or `while` loop since while we do not support IE 8 officially, the following points is: 1) we should also not break the code execution in the browser 2) we should not limit the usage that MUST be above IE 8.

letchoo:
Closing because if users want to integrate themselves with no guarantees, they can append the polyfill themselves before the `<script src="adapter.debug.js"></script>`.

PR #288, opened 2017-09-14
AJS-293: Prevent listening to messages that does not come from self to extension

letchoo:
This resolves the issue where `getUserMedia({ video: { mediaSource: "screen" } })` for chrome as it listens to all messages in the same page. Currently in the deprecated method which some apps would still be using, the extension would listen to messages from `window.postMessage()` and react upon it. That however is a current problem in the AdapterJS since it is listening to all messages instead of acting upon specific messages, and that causes issues for JS plugins that are using the `window.postMessage()` as well. 

The current PR just simply listens for specific messages instead.

### DIFF
--- a/source/adapter.screenshare.js
+++ b/source/adapter.screenshare.js
@@ -377,16 +377,11 @@ AdapterJS._defineMediaSourcePolyfill = function () {
       // Listen to iframe messages
       var getSourceIdFromIFrame = function (sources, cb) {
         window.addEventListener('message', function iframeListener (evt) {
-          // Unload since it should be replied once if success or failure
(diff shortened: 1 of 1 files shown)


johache:
LGTM, I just wanna make sure that the logic is : 
if evt.data.chromeExtensionStatus === whatever, remove the event handle and do something
else, ignore the message **and keep the event handler**

letchoo:
@johache yes that is the logic intended, so that we don't remove the event listener when it shouldn't.

letchoo:
@johache so it's considered mergeable?

johache:
Yep, I already approved the PR :)

PR #291, opened 2017-09-29
Add bind() polyfill to plugin functions

jeffstephens:
The Temasys plugin returns Javascript functions that reflect a much older version of Javascript and do not support the `bind()` functionality which causes failures in our code and some plugins like [rtcpeerconnection](https://github.com/otalk/RTCPeerConnection). This recursively polyfills all functions on the plugin object so they support this relatively recent Javascript functionality.

### DIFF
--- a/source/adapter.js
+++ b/source/adapter.js
@@ -382,6 +382,23 @@ AdapterJS.addEvent = function(elem, evnt, func) {
   }
 };
 
+AdapterJS.recursivePolyfillBind(obj) {
+  if (!obj) {
+    return;
+  }
+
+  var pluginProperties = Object.keys(obj);
+  for (var i = 0; i < pluginProperties.length; i++) {
+    var currentProperty = obj[pluginProperties[i]];
+
+    if (typeof currentProperty === 'function' && !currentProperty.bind) {
+      currentProperty.bind = Function.prototype.bind;
+    } else if(typeof currentProperty === 'object') {
+      AdapterJS.recursivePolyfillBind(currentProperty);
+    }
+  }
+}
+
 AdapterJS.renderNotificationBar = function (message, buttonText, buttonCallback) {
   // only inject once the page is ready
   if (document.readyState !== 'interactive' && document.readyState !== 'complete') {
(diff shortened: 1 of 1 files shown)


johache:
Will need to test, but in theory, it sounds great to me :)

jeffstephens:
Sweet! I do believe I'm seeing infinite recursion in IE11 due to the different event listener code (`addEvent`)... I'll fix that up.

Also, would it be better if I squashed these commits into one?

johache:
You don't need to squash the commits.

I will be waiting for an update on the infinite recursion :)

jeffstephens:
Ok, that should do it! Let me know if you'd like me to make any changes.

johache:
Hi,
I'm afraid this is not working as intended.

# Safari
Take for example this sample : https://plugin.temasys.com.sg/demo/src/content/peerconnection/pc1/
You can modify it to use your AJS.

Try the following code : 
```var rs = pc2.getRemoteStreams;
rs.bind(pc1)(); // returns pc2.getRemoteStreams(), instead of pc1.getRemoteStreams()```

This is probably due to the nature of NPAPI/ActiveX plugin. When you call pc.createOffer, it's not calling the createOffer function on the pc1 instance. It's calling **the** createOffer function of this specific instance of PC. pc2.createOffer is a difference instance of the function.
I know it's a bit weird, but I'm pretty sure that's what's happening.

# IE
IE is flat out not working for me. I tried to bind pc.createOffer and I am getting an immediate error when I try to call a function after binding it.

**Comment by jeffstephens** [comment]
Whoops - thanks for the info, that is helpful. The plugin objects are a bit strange. I'll work on getting this fixed!

**Comment by johache** [comment]
Feel free to work on it as much as you want, but very honestly, I don't think that there is a clean way of fixing this.

**Comment by jeffstephens** [comment]
Thanks for the help! I think you are probably right and I'm going to start looking at a different approach than modifying AdapterJS.

**Comment by jeffstephens** [comment]
Hi @johache - my colleague @dwilson6 figured out a way to make this work with the plugin objects. I'm now able to call `bind()` on the functions as in Chrome/Firefox. Would you mind taking another look?

**Comment by johache** [comment]
Sure, no prob.
I'll either get to it today/tomorrow, or it will have a to wait for a couple of weeks.

**Comment by jeffstephens** [comment]
Hey @johache, do you have time to look at this? Thanks for your attention!

**Comment by johache** [comment]
I'm sorry, but it still doesn't seem to work.

```
var rs = pc2.getRemoteStreams;
rs.bind(pc2)(); // same error for rs.bind(pc2)();
```

Gives me this error
```TypeError: rs.bind(pc2) is not a function. (In 'rs.bind(pc2)()', 'rs.bind(pc2)' is an instance of Array)```

and trying to preset actually calls the function :
``` 
var co = pc1.createOffer.bind(onCreateOfferSuccess, onCreateSessionDescriptionError,
      offerOptions);
```

ee11131:
In my case, when I merge this pull request in adapter, using the latest simplewebrtc I get the error "Calling method on NPObject"  on "`return fToBind.apply(this instanceof fNOP`". Note that SimpleWebRTC uses bind() so that is why it's so important to me to get this pull request to work.

PR #295, opened 2017-12-08
AJS-304 Strict mode

johache:
AJS written in strict mode.
Many variables declared with var, some scoped into AJS


letchoo:
Seems good so far, but will need to test on IE 9/10/11/12 for sanity reasons.

johache:
Tested ok against IE 9, IE 11, and IE 11 in IE 10 emulator mode (I don't have an actual IE 10)

PR #300, opened 2018-01-19
Detect global AdapterJS object from Window

### LINKED ISSUE #299: AdapterJS global object is not properly detected when used with module bundlers like WebPack
We cant properly setup global global AdapterJS options like `forceSafariPlugin` when including library using module bundler (i.e. WebPack).

Consider this code:

```
window.AdapterJS = {
    options: {
        forceSafariPlugin: true
    }
};
require('adapterjs');
```
The option is actually not being set because in library you dont explicity try to grab `AdapterJS` object from Window:
```
var AdapterJS = AdapterJS || {};
```
 and bundlers like webpack resolves such runtime require with functions 
that adds scope to execution and this kind of syntax is not working.

Super simple solution for this is to use this kind of syntax: 
```
var AdapterJS = AdapterJS || window.AdapterJS || {};
```

Without this change i actually dont know how to override those options in such case.

Pull Request created: #300

aszmyd:
Fixes #299


johache:
@letchoo I'm happy with this PR.
@aszmyd Could you just change it to go into dev. Thx :)

aszmyd:
@johache sorry. done

johache:
@aszmyd no worries. Thanks for the correction.
Like I wrote in my previous message, I just want @letchoo to review as well. (due process :) )

aszmyd:
@johache thanks. do you have any schedule when i can expect it to be released?

johache:
We don't have an immediate release date right now. Probably within 3 weeks.

PR #305, opened 2018-09-13
Added check for navigator.mimeTypes undefined. Should show plugin not supported banner

ishannaithani:
**Purpose of this PR**

In situations where navigator.mimeTypes are not available (headless browsers), the AJS code needs to make a safe check and display plugin not supported banner.


johache:
``notInstalledCb`` will trigger the installation of the plugin. That's not necessary in the case of mimetype being undefined. The notification is enough.

ishannaithani:
Understood. I have remove the call to `notInstalledCb`.

PR #306, opened 2018-11-15
Ajs 309: when use JSDom, variables are not attached to any scope

leonardotemasys:


ishannaithani:
Probably would want to commit into development first rather than master?

leonardotemasys:
Changed to dev

PR #311, opened 2019-11-15
Surangap

surangap:
Fixed plugin autoupdate popup | exposed utils.addEvent via AdapterJS

### DIFF
--- a/src/plugin/plugin_manager.js
+++ b/src/plugin/plugin_manager.js
@@ -77,17 +77,6 @@ function maybeThroughWebRTCReady() {
   });
 };
 
-function checkUpdateAvailable(resolve_callback, reject_callback) {
-  if (pluginObject) {
-    let current = pluginObject.VERSION;
-    let latest  = getLatestVersionNumber();
-    (utils.versionCompare(latest, current) > 0) ? resolve_callback(true) : resolve_callback(false);    
-
-  } else {
-    reject_callback("plugin object is not available");
-  }
-}
-
 ////////////////////////////////////////////////////////////////////////////
 /// 
 /// exports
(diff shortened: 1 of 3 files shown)


src/plugin/plugin_manager.js
```diff
@@ -174,10 +174,23 @@ export function getLatestVersionNumber() {
...
-  if (!pluginObject) return;
-  let current = pluginObject.VERSION;
-  let latest  = getLatestVersionNumber();
-  return utils.versionCompare(latest, current) > 0;
+  return new Promise ((resolve, reject) =>{
+    callWhenPluginReady(() => {
+      if (pluginObject) {
+        var current = pluginObject.VERSION;
+        var latest = getLatestVersionNumber();
+        utils.versionCompare(latest, current) > 0 ? resolve(true) : resolve(false);
```
johache: would read nicer as 
resolve(utils.versionCompare(latest, current) > 0);

src/plugin/plugin_manager.js
```diff
@@ -174,10 +174,23 @@ export function getLatestVersionNumber() {
...
+      } else {
+        reject("plugin object is not available");
+      }
+    });
+  });
+}
+
+export function checkForUpdate() {
```
johache: I don't think we need to keep this function.
At most, you could have a updateIfNewVersionAvailable() function, which would NOT depend on config.autoUpdate, and would just be 
if (isUpdateAvailable()) {
      updatePlugin();
    }

johache (changes requested):

src/plugin/plugin_manager.js
```diff
@@ -174,10 +174,23 @@ export function getLatestVersionNumber() {
...
+  if (isUpdateAvailable() && config.autoUpdate) {
```
miniruwan: Better to change the order of the conditions; Do the config.autoUpdate check first and then do the heavy isUpdateAvailable() check (which involves http GET).

miniruwan (changes requested):

src/plugin/plugin_manager.js
surangap: function removed since no longer use

